PR #2, opened 2016-09-21
simplify code to construct output-filename

haraldp271:
Harald is curious how pull-requests work.  As a by-product, simplify the code that constructs the output filename.  Mot sure why the original version used w_m.descriptor_string.  It strikes me that this does either nothing, or it causes the output filename to be a "_CoM.h5".

### DIFF
--- a/SpEC/com_motion.py
+++ b/SpEC/com_motion.py
@@ -209,7 +209,17 @@ def remove_avg_com_motion(path_to_waveform_h5='rhOverM_Asymptotic_GeometricUnits
     w_m = read_from_h5(path_to_waveform_h5)
 
     # Output filename: insert '_CoM' into the original `path_to_waveform_h5`
-    path_to_new_waveform_h5 = path_to_waveform_h5.replace('.h5', '_CoM.h5')
+    if path_to_waveform_h5.count(".h5")>1:
+        print "The path ro the waveform contains '.h5' multiple times."
+        print "This is bad practice, and makes it unclear how to produce"
+        print "an output path".
+        os.sys.exit(-1)
+    tmp = path_to_waveform_h5.replace('.h5', '_CoM.h5')
+
+    # remove the descriptor string from the input-path, as it is added
+    # back automatically in write_waveform
+    path_to_new_waveform_h5 = tmp.replace(w_m.descriptor_string+'_','')
+
 
     # Get the CoM motion from Horizons.h5
     x_0,v_0,t_0,t_f = estimate_avg_com_motion(path_to_horizons_h5=path_to_horizons_h5,
--- a/SpEC/com_motion.py
+++ b/SpEC/com_motion.py
@@ -208,10 +208,8 @@ def remove_avg_com_motion(path_to_waveform_h5='rhOverM_Asymptotic_GeometricUnits
     # Read the waveform data in
     w_m = read_from_h5(path_to_waveform_h5)
 
-    # Insert '_CoM' into the original `path_to_waveform_h5`; this will be the output path
-    path_to_new_waveform_h5 = path_to_waveform_h5.split('.h5', 1)
-    path_to_new_waveform_h5.insert(1, '_CoM.h5')
-    path_to_new_waveform_h5 = ''.join(path_to_new_waveform_h5).replace(w_m.descriptor_string + '_', '')
+    # Output filename: insert '_CoM' into the original `path_to_waveform_h5`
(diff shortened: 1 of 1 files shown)


SpEC/com_motion.py
```diff
@@ -208,10 +208,8 @@ def remove_avg_com_motion(path_to_waveform_h5='rhOverM_Asymptotic_GeometricUnits
     # Read the waveform data in
     w_m = read_from_h5(path_to_waveform_h5)
 
-    # Insert '_CoM' into the original `path_to_waveform_h5`; this will be the output path
-    path_to_new_waveform_h5 = path_to_waveform_h5.split('.h5', 1)
-    path_to_new_waveform_h5.insert(1, '_CoM.h5')
-    path_to_new_waveform_h5 = ''.join(path_to_new_waveform_h5).replace(w_m.descriptor_string + '_', '')
```
moble: **Explanation:**  The `descriptor_string` is automatically computed and prepended to the filename in `write_to_h5`, so it needs to be removed here in order to not have filenames that start with, e.g., `rhOverM_rhOverM_`.  This serves as a bit of a check that everything is working properly, because the output file will only be named correctly if the input file was correctly interpreted.  That's important because the transformation depends on the data type — for example, if the data is `h` or `Psi4`, if `r` is scaled out or not, and if `M` is scaled out or not.  So an unexpected filename is the big warning that something messed up, and should be very obvious to the user even if s/he ignored all the stdout while the code was running.

SpEC/com_motion.py
```diff
@@ -208,10 +208,8 @@ def remove_avg_com_motion(path_to_waveform_h5='rhOverM_Asymptotic_GeometricUnits
...
+    # Output filename: insert '_CoM' into the original `path_to_waveform_h5`
+    path_to_new_waveform_h5 = path_to_waveform_h5.replace('.h5', '_CoM.h5')
```
moble: **Action item:** Please add `, 1` to the end of the `replace` argument, and tack on `.replace(w_m.descriptor_string + '_', '')`.

SpEC/com_motion.py
```diff
@@ -208,10 +208,8 @@ def remove_avg_com_motion(path_to_waveform_h5='rhOverM_Asymptotic_GeometricUnits
     # Read the waveform data in
     w_m = read_from_h5(path_to_waveform_h5)
 
-    # Insert '_CoM' into the original `path_to_waveform_h5`; this will be the output path
-    path_to_new_waveform_h5 = path_to_waveform_h5.split('.h5', 1)
-    path_to_new_waveform_h5.insert(1, '_CoM.h5')
```
haraldp271: With this explanation, the original code makes sense.  However, to improve code-clarity, it is useful to make assumptions explicit.  So, instead of adding an opaque ",1" in an already
opaque replace statement, I added a check that there is only **one** .h5 within the entire path; if not an error is printed.  

NOTE: I don't know the error reporting strategy of scri, so this is a (likely) unacceptable print statement, followed with sys.os.exit().

SpEC/com_motion.py
haraldp271: If a piece of code cannot handle input data, I would hope such code will generate a clear error message (or just fail).  As such, using the output filename as the indicator of success seems unintuitive to me, especially, if there is no other warning nearby that this is happening.

What I found particularly confusing were implict assumptions of the code.  Specifically, the dependency of descriptor_string on the file-name.  (I wondered 'what if the descriptor_string is NOT contained in the filename?  Or what if somebody decides to modify the descriptor_string, e.g. making it more verbose?').

Anyway, replace(descriptor_string,...) added back.

SpEC/com_motion.py
moble: I don't want it to error out, because it should still work if someone uses such a path; I just want to change the first occurrence of the `.h5` string.  (But for future reference, I just use standard exceptions like `raise ValueError("Some message")`.)

SpEC/com_motion.py
moble: But it *probably can* handle the input data.  There's no way to tell if the user incorrectly named a file and thus wanted a different type of transformation, or if the user correctly named the file and wanted the type of transformation that `scri` gave it.  If the output file is named in some unexpected way, the user should notice -- despite the fact that there's no way for `scri` to know what s/he wanted.

moble:
Superseded by a7fb3f8

PR #3, opened 2016-11-09
Adding output file name to SpEC/com_motion.py function remove_avg_com_motion()

cwoodford:
To allow the user to name and put their output COM corrected h5 file whereever they want. The remove_avg_com_motion() function can take "outfile" as an argument and name the output file accordingly. If nothing is specified, the default value is "None" and names the output file the same way the original function would.

I don't think this causes any major disturbances. I tested it and it seems to work.


moble:
Two things.  First, byte code files `*.pyc` should never be added to a repo.

Second, why do you want this change?  From my earliest days of writing waveform code, I've found that many mistakes were being made because waveform files all look the same (a bunch of numbers), even though the code has to know what those numbers represent (mass scaled out; radius scaled out; h or Psi4; etc.) in order to correctly manipulate them.  People would pass incorrect inputs, or try operations that don't make sense with a given data type.  Far and away, the best way I've found to deal with this problem is to compute the name, using what the code already knows about the data types.  By computing the name, people are forced to look at what the code thought the data actually represented

Is there something you don't like about the names coming out of the code?  If there's something wrong about the output, I'd prefer that we fix it.  If it's just a question of aesthetics, I'd prefer that the output just be renamed outside of this function, so that whatever wrapper is calling it has to acknowledge what's actually happening at some point.  What do you think?

cwoodford:
That's my bad for the pyc files, I hadn't intended to add them.

I don't think there's anything really wrong with the naming conventions, this was something that Harald and I had discussed previously and I'm only getting around to now. We were thinking that it would be more convenient to be able to direct the function where you want the output and if you want it named something different. However, this does fall under aesthetics. If this was a problem before, like you mentioned, that it might be best to leave things the way they are.

PR #6, opened 2019-02-20
Allow com correction function to copy over VersionHist.ver

10220:
This allows the scripts in CatalogAnalysis that use `scri.SpEC.remove_avg_com_motion` to copy the `VersionHist.ver` from the source file to the new file.


10220:
This functionality will be moved exclusively to CatalogAnalysis/COMCorrection.py and CatalogAnalysis/COMCorrectionRWZonly.py

PR #7, opened 2019-03-21
Support extrapolation on all Weyl scalars

10220:
This does a few things:
- Fixes a bug in the extrapolation procedure (data would have NaN's) 
- Fixes a bug where output wouldn't be printed to non-NRAR format 
- Allows extrapolation to support all Weyl scalars
- Chooses default Lmin based on DataType (i.e. spin-weight)

### DIFF
--- a/SpEC/file_io.py
+++ b/SpEC/file_io.py
@@ -242,9 +242,9 @@ def read_from_h5(file_name, **kwargs):
     return w
 
 
-def write_to_h5(w, file_name, file_write_mode='w', attributes={}):
+def write_to_h5(w, file_name, file_write_mode='w', attributes={}, use_NRAR_format=True):
     """
-    Output the Waveform in NRAR format.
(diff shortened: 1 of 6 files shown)


10220:
Hey @moble, everything should be good now. Scri's extrapolation procedure supports VersionHist.ver with the newest commit I added to this PR.

markscheel:
@moble I would like to change the SpEC postprocessing scripts to use scri instead of GWFrames for extrapolation.  Then the scripts (especially a future version that will run as a cron job) will be simpler and will have fewer dependencies. Is there any reason not to do this?  If we want to move from GWFrames to scri, I think we need this PR or some modified version of it.

10220:
Hey @moble, let me know if there is still any other changes you wanted. Otherwise at this point I think this is all set.

10220:
Also both conda-forge and pip have the updated versions of quaternion.

PR #8, opened 2019-05-16
Create complex inner_prod between timeseries, and WaveformMode objects

duetosymmetry:


waveform_modes.py
```diff
@@ -315,6 +315,29 @@ def indices(self, ell_m):
             raise ValueError("Input `ell_m` should be an Nx2 sequence of integers")
         return ell_m[:, 0] * (ell_m[:, 0] + 1) - self.ell_min ** 2 + ell_m[:, 1]
 
+
+    def inner_prod(self, b):
```
10220: For clarity, would it be better to have a different name for the inner product here and the one in `mode_calculations.py`?

waveform_modes.py
```diff
@@ -315,6 +315,29 @@ def indices(self, ell_m):
...
+            The other set of modes to inner product with.
+
+        Returns
+        -------
+        inner_prod : complex
+            <self, b>
+        """
+
+        # QUESTION/TODO: Should we do validation here?
+        # Probably? Ensure that times and LMs are the same, not just same shape!
```
10220: Perhaps leave it as an option with an argument that the user can turn on and off?

mode_calculations.py
```diff
@@ -452,3 +452,48 @@ def corotating_frame(W, R0=quaternion.one, tolerance=1e-12, z_alignment_region=N
...
+    from scipy.integrate
+
+    Parameters
+    ----------
+    t : array_like
+        Time samples for waveforms astar and b.
+    astar : array_like
+        The conjugate of the 'a' waveform in the inner product (or
+        simply a, if conjugate_a=True has been passed).  Must have the
+        same shape as b.
```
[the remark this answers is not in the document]
duetosymmetry: Is the wording bad? Currently it mentions `conjugate_a`, but maybe the parenthetical is confusing?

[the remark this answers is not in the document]
10220: I guess the name `astar` threw me off at first. Eh, disregard my comment actually. It's all good :+1:

waveform_base.py
```diff
@@ -818,6 +818,29 @@ def interpolate(self, tprime):
         return W
 
 
+    @waveform_alterations
+    def conjugate(self):
+        """Complex conjugate the waveform data
```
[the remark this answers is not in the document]
duetosymmetry: Good point that there is something subtle here. This method gets inherited by both WaveformModes and WaveformGrid, and has a different action in the two cases: either conjugating w(u,theta,phi) or w_lm(u). Both operations may be desired, but they are distinct and should not be confused. Maybe the name is bad for WaveformModes. Maybe WaveformModes also needs a function which applies conjugation, then sends the mode (l,m) mode to (l,-m) and multiplies by the appropriate parity?

waveform_modes.py
duetosymmetry: I like this idea.

mode_calculations.py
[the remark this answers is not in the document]
moble: Not a very strong objection, but my convention (in this package and elsewhere) is to use bar for conjugation, especially as it leaves room for mode indices.  Maybe being more explicit would help, as in `a_conjugate` and `apply_conjugate` or something?

waveform_base.py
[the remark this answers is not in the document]
moble: Is this function actually needed?  I'm a little worried because it breaks certain promises made by the code, related to the fact that a Waveform is supposed to know what its data type is (which is how it knows its spin weight, boost weight, etc.).  So if you really want this function, we'd need to add each data type's conjugate to the list...

waveform_modes.py
moble: I don't like the idea of turning it off.  The validation should be quick, so there would be no good reason to skip it.  Moreover, I expect that it would be reasonable to automatically do some interpolation to the intersection of the two timeseries, and similar things, rather than just checking.  Maybe an option to issue warnings rather than raise — for example if the spin weights are not equal, or if some LM modes are not present in one.

waveform_modes.py
moble: I'd be in favor of actually calling this `inner_product`; the missing letters make me worry something bad happened to Leo just as he was typing the function name.  Or maybe even `__matmul__`!  :)

waveform_base.py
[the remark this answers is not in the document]
duetosymmetry: @moble good point, I didn't think more deeply about that fact that a conjugated waveform should have opposite spin weight. I now think this is too subtle an issue to be handled by a function at the level of WaveformBase and will remove it (since also it is not used by the rest of this PR)

waveform_modes.py
duetosymmetry: Something terrible happened to me—I felt like a fortran programmer and started abbreviating things instead of being verbose!

Not sure I understand the suggestion/reference to __matmul__. The timesteps must also be taken into account as weights for the sum. Can you clarify?

waveform_modes.py
duetosymmetry: Thanks for pointing out the spin as well. So validation should be default.

Different s's: fail

Different set of (ell,m)'s: could fail, or only use the set of (ell,m)'s in the intersection of the two waveforms. Should this be an option?

Different set of times: could fail, or interpolate to a common set of times on the intersection between the two waveforms. But I'm not sure what timestep should you use if the timesteps differ. Thoughts?

waveform_modes.py
moble: > I felt like a fortran programmer

Eh, that's still better than Mark, who programs everything as if it were Perl.

If you `def __matmul__(self, b)`, you'd be able to call it as `a @ b`.  The function would still have access to all of the information in `a` and `b` (including the timesteps), but would just make you feel cooler when you call it.

Of course, since so much of my "documentation" relies on being able to find functions by way of tab completion, it's probably better to stick with an actual name...

waveform_modes.py
moble: > Different set of (ell,m)'s: could fail, or only use the set of (ell,m)'s in the intersection of the two waveforms. Should this be an option?

I'd say at least a warning would be appropriate, but I think it should be allowed, because IIRC the output from `WaveformModes.transform` will have a larger ell_max than the input [as it should, whenever there's a boost or (super)translation].

> But I'm not sure what timestep should you use if the timesteps differ.

The [method I created for extrapolation](https://github.com/moble/scri/blob/5875404aceb42905ad2d387278c3341d9c2e6d10/extrapolation.py#L48-L64) is not unique, but I think it's sensible, and it's been tested over time.

mode_calculations.py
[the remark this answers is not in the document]
duetosymmetry: Renamed to abar, and changed to apply_conjugate for clarity.

waveform_base.py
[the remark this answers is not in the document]
duetosymmetry: Removed.

waveform_modes.py
duetosymmetry: I did not know about the @ magic or any other magic operators. I think of waveforms as vectors, not matrices, so I don't like the semantics of using @ for this purpose. However, support the idea of overloading addition/subtraction/multiplication by scalar (real or complex) for waveforms.

waveform_modes.py
duetosymmetry: For now I updated the PR to validate equality spins, LMs, and times. The argument for intersections of LMs and times is good. I propose that the default behavior is require everything matches, unless an optional parameter is passed saying "I know what I'm doing" and to use the intersections of LMs and times — because there may be interpolation which is a potentially expensive operation. Thoughts?

10220: Given the conversation I like the idea of having the default be the validation, but it's always good to allow the user to skip it. I don't imagine the interpolation would be too costly, but regardless its good to allow a user to skip it for a rare case.

10220: I definitely agree with calling this `inner_product` but then I do propose we change the name of the `inner_product` function in `mode_calculations`. I think having both of them with the same name will be too confusing.

waveform_modes.py
moble: > I think of waveforms as vectors, not matrices

Vectors that *aren't* matrices?!  The `@` operator can handle both, of course.  Anyway, for reasons explained above, I'm happy with this version.

As for overloading scalar operators, I implemented them in GWFrames, but eventually came to the conclusion that it's not actually useful because the action to be taken is ambiguous, and can either already be done easily and with less ambiguity (e.g., `h.data *= 3.14`), or should be given its own function (e.g., rescaling mass or distance).

waveform_modes.py
duetosymmetry: Yeah, it's not just a code issue but a mathematics issue: lots of vector spaces have inner products, and there might even be different inner products on the same vector space, but we rarely give them unique names — typically <a,b> and you have to know from context that the operation < , > is on VxV->C, or WxW->C, etc.

One of these is the inner product on the space of timeseries R->C (or some other V->C), and the other is the inner product on the space of S^2xR -> C.

Is it clearer if we call one timeseries_inner_product, and the other scri_inner_product?

I would say that, similar to mathematics usage, we do have context whenever one of these methods is invoked. One of them is a member of WaveformModes and takes as argument a second wm, while the other method takes two vectors.

moble: There are two different types of "validation" happening here, and you should never be able to *skip* either one.  The first type is checking for things that can be handled but will affect the code path — like different time steps or modes.  In this case, the option should be between a default of printing a warning but doing what needs to be done, or just doing what needs to be done.  The second type is something like checking the spin weights; the "inner product" of functions with different spin weights will be a coordinate-dependent thing, and thus pretty meaningless.  We therefore want to strongly discourage people from doing this.  In this case, the option could be between raising an error or printing a warning.  If the user doesn't like seeing the warning s/he is free to use `warnings.catch_warnings` or `contextlib.redirect_stderr`, but we want to impose that penalty because the cost of silence will be bad physics.

duetosymmetry: Am now validating, and have introduced options allow_LM_differ and allow_times_differ. These will respectively slice (in ell, m) and/or interpolate (in time) if needed to perform the overlap.

duetosymmetry: Note, this PR is blocked by #9 if you try to both slice in LM and interpolate in time.

10220: > Is it clearer if we call one timeseries_inner_product, and the other scri_inner_product?
I think this would be better

duetosymmetry: I just rebased on top of master after #10 was merged, so this is no longer blocked by the issue. I think the only remaining issue is names for the two different methods, unless we want to consider whether scipy.integrate.simps is the correct approach or not.

waveform_modes.py
```diff
@@ -315,6 +315,77 @@ def indices(self, ell_m):
...
+                                 min(self.ell_max, b.ell_max) + 1 )
+                if (LM_clip.start >= LM_clip.stop):
+                    raise ValueError("Intersection of (ell,m) modes is "
+                                     "empty.  Assuming this is not desired.")
+            else:
+                raise ValueError("ell_min and ell_max must match in inner_product "
+                                 "(use allow_LM_differ=True to override)")
+
+        t_clip = None
+        if ((self.t.shape != b.t.shape) or (self.t != b.t).any()):
```
moble: Just FYI: this is basically what `np.array_equal` does.

waveform_modes.py
moble: I don't expect that it would make a huge difference numerically, but I would favor replacing `simps` with a call to `InterpolatedUnivariateSpline.integral(t1, t2)`.  On that note, it seems like adding arguments named something like `t1=None, t2=None` would make sense, if people want to cut out junk radiation or ringdown, for example.  These can be passed directly to `intersection`.

moble: On a related note, I would have thought that the most efficient way would be to do the sum over modes first, and then the integration over time.  Especially if you switch to spline integration, splines are so slow (and annoying to call) that this would surely improve the speed dramatically.  Thoughts?

duetosymmetry: Good point about summing over modes first. I should have thought of that.
Let me investigate `InterpolatedUnivariateSpline.integral` in a bit...

duetosymmetry: `InterpolatedUnivariateSpline` is more annoying for the basic `inner_product`. As it is written now, mode_calculations.py:inner_product will happily integrate along some axis of an ndarray. Using `InterpolatedUnivariateSpline`, we would need to build k interpolants, where k is the product of the # of elements in the axes that are not being integrated. Maybe that function should just be restricted to operating on vectors, rather than ndarrays?

duetosymmetry: `InterpolatedUnivariateSpline` does not seem to be too much slower than `simps`. Now using it in `WaveformModes.inner_product`, but haven't decided what to do for the more general `mode_calculations.py:inner_product`.

waveform_modes.py
duetosymmetry: Done.

Issue #9, opened 2019-05-17
Complex Waveform that has been sliced in (ell,m) can not be interpolated in time


duetosymmetry:
Because scipy.interpolate functions only act on floats, not complex, WaveformBase.interpolate contains the following code to go through real views:

https://github.com/moble/scri/blob/72aaa1aa8ea4a86b4435f2a6bcef95d02a31ac97/waveform_base.py#L807-L810

However, if the Waveform was built by slicing a WaveformModes in (ell,m), then data_2d will not be C-contiguous. Then np.ndarray.view will raise a ValueError, [as mentioned here](https://www.numpy.org/devdocs/reference/generated/numpy.ndarray.view.html).

duetosymmetry:
What is the penalty for doing:

```
        if self.data.dtype == np.dtype(complex):
            for i in range(W.n_data_sets):
                W.data_2d[:, i] = (splev(W.t, splrep(self.t, self.data_2d[:, i].real, s=0), der=0)
                                   + 1j*splev(W.t, splrep(self.t, self.data_2d[:, i].imag, s=0), der=0))
```

which avoids the error?

duetosymmetry:
I have confirmed that with the version of numpy that I have installed (1.16.3 or later), `.real` and `.imag` make shallow copies, even with sliced arrays. It is rather easy to check. Below is an example session showing this. This is also a minimal working example for the bug above and the fix in #10.
Therefore there is no penalty for using `.real` and `.imag`.

```python
>>> a = np.array([[1., 2., 3.],[7.,8.,9.]]) + np.array([[4., 5., 6.],[10.,11.,12.]])*1.j
>>> print(a)
[[1. +4.j 2. +5.j 3. +6.j]
 [7.+10.j 8.+11.j 9.+12.j]]
>>> d = a[:, 0:2]
>>> print(d)
[[1. +4.j 2. +5.j]
 [7.+10.j 8.+11.j]]
>>> d.base is a
True
>>> d.data.c_contiguous
False
>>> dv = d.view(dtype=float)
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
<ipython-input-83-eaec6c6460fd> in <module>
----> 1 dv = d.view(dtype=float)

ValueError: To change to a dtype of a different size, the array must be C-contiguous
>>> dr = d.real
>>> di = d.imag
>>> (dr.base is a) and (di.base is a)
True
>>> print(di)
[[ 4.  5.]
 [10. 11.]]
[...]

PR #10, opened 2019-05-17
Use .real and .imag instead of .view(dtype=float) in case of slicing

duetosymmetry:
Bugfix for #9. np.ndarray.view will raise a ValueError if data_2d is not C-contiguous, [as mentioned here](https://www.numpy.org/devdocs/reference/generated/numpy.ndarray.view.html). Using .real and .imag seems more robust. ~I don't know if this always does a deep copy, but I hope that python will detect if a shallow copy is possible~. EDIT: There is no deep copy, see [this example](https://github.com/moble/scri/issues/9#issuecomment-494516167)


moble:
Closes #9

PR #13, opened 2019-05-24
WIP: Add function to compute momentum flux

moble:
As suggested in https://github.com/sxs-collaboration/CatalogAnalysis/issues/23, I've started adding code to compute the linear momentum flux from a waveform.  At the moment, this includes a pretty rigorous test, except that the test references waveform files on my laptop; I need to create some good analytic waveforms to really test this properly.

I should also add a convenience function to integrate this over some period of time, and the very similar code needed to compute the energy and angular momenta.


moble:
Closed in favor of #16, which incorporates everything here and more.

PR #14, opened 2019-06-05
General inner product functionality in mode_calculations

duetosymmetry:
This is baseline code which uses Simpson's rule from scipy.integrate.simps for quadrature. A better approach would be to use interpolation, but this will require using a wrapper function that @moble wrote for `quaternion`

### DIFF
--- a/__init__.py
+++ b/__init__.py
@@ -86,6 +86,7 @@ DataNamesLaTeX = [r"\mathrm{unknown data type}", r"\psi_0", r"\psi_1", r"\psi_2"
 # Set up the WaveformModes object, by adding some methods
 from .waveform_modes import WaveformModes
 from .mode_calculations import (LdtVector, LVector, LLComparisonMatrix, LLMatrix,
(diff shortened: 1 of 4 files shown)


duetosymmetry:
@moble is [`quaternion.spline_definite_integral`](https://github.com/moble/quaternion/blob/5804b4085ccbdceab8f097cf14f45d126e708489/calculus.py#L323) now available in the PyPI version of quaternion? Should I rewrite this PR to use those functions?

moble:
It is.  I just realized that I had hard-coded the axis to 0, for some reason, but I've now fixed that.  So as long as your version is today's, you should be able to use it as a drop-in replacement for `simps`.

duetosymmetry:
The general inner_product function and the one in waveform_modes both use `quaternion.calculus.spline_definite_integral` now. Since `scipy` is a requirement for this package, we are guaranteed that `quaternion` will be able to use InterpolatedUnivariateSpline.

PR #16, opened 2019-07-26
Add fluxes of E, P, J, via a general matrix-expectation-value function

duetosymmetry:
All three of these have expressions in terms of quadratic combinations of the modes of h, which sandwich some linear operators. All those operators are sparse in the SWSH basis, so they can be computed in O(N) where N is the number of modes. This is accomplished with a higher order function `matrix_expect_val(a, M, b)`, which computes <a|M|b>. The matrix M is defined in terms of a generator `M(ell_min, ell_max)` which is then iterable, yielding tuples `(ellp, mp, ell, m, M_el)` that correspond to non-vanishing matrix elements <ellp, mp|M|ell, m>.

### DIFF
--- a/mode_calculations.py
+++ b/mode_calculations.py
@@ -600,3 +600,167 @@ def inner_product(t, abar, b, axis=None, apply_conjugate=False):
         integrand = np.conjugate(abar)*b
 
     return sdi(integrand, t, axis=axis)
+
+
+def matrix_expect_val(a, M, b,
+                      allow_LM_differ=False, allow_times_differ=False):
+    """The matrix expectation value <a|M|b>(u), where M is a linear operator.
+
+    Treat two spin-s waveforms a, b (in the modal representation) as vectors.
+    Then we can form the 'expectation value'
+
+    .. math:: <a|M|b>(u) \equiv \sum_{\ell', m', \ell, m} a^*_{\ell' m'}(u) M_{\ell' m' \ell m} b_{\ell m}(u) \,.
+
+    For efficiency with sparse matrices, M should be implemented as a
+    generator which yields tuples (ellp, mp, ell, m, M_{ellp mp ell m})
+    for only the non-vanishing matrix elements.
+
+    Parameters
+    ----------
+    a : WaveformModes object
+        The waveform |a>. This function is "antilinear" in `a`.
+
+    M : evaluable
+        M will be called like
+        `for ellp, mp, ell, m, M_el in M(ell_min, ell_max):`
+        This is best implemented as a generator yielding tuples
+        `(ellp, mp, ell, m, M_{ellp mp ell m})`
+        which range over ell, ellp values up to and including ell_max,
+        for only the non-vanishing matrix elements of M.
+
+    b : WaveformModes object
+        The waveform |b>. This function is linear in `b`.
+
+    allow_LM_differ : bool, optional [default: False]
+        If True and if the set of (ell,m) modes between a and b
+        differ, then the inner product will be computed using the
+        intersection of the set of modes.
+
+    allow_times_differ: bool, optional [default: False]
+        If True and if the set of times between a and b differ,
+        then both WaveformModes will be interpolated to the
+        intersection of the set of times.
+
+    Returns
+    -------
+    times, expect_val : ndarray, complex ndarray
+        Resulting tuple has length two.  The first element is the set
+        of times u for the timeseries.  The second element is the
+        timeseries of <a|M|b>(u).
+
+    """
+
+    from .extrapolation import intersection
+    from spherical_functions import LM_index
+
+    if (a.spin_weight != b.spin_weight):
+        raise ValueError("Spin weights must match in matrix_expect_val")
+
+    LM_clip = slice(a.ell_min, a.ell_max + 1)
+    if ((a.ell_min != b.ell_min) or (a.ell_max != b.ell_max)):
+        if (allow_LM_differ):
+            LM_clip = slice( max(a.ell_min, b.ell_min),
+                             min(a.ell_max, b.ell_max) + 1 )
+            if (LM_clip.start >= LM_clip.stop):
+                raise ValueError("Intersection of (ell,m) modes is "
+                                 "empty.  Assuming this is not desired.")
+        else:
+            raise ValueError("ell_min and ell_max must match in matrix_expect_val "
+                             "(use allow_LM_differ=True to override)")
+
+    t_clip = None
+    if not np.array_equal(a.t, b.t):
+        if (allow_times_differ):
+            t_clip = intersection(a.t, b.t)
+        else:
+            raise ValueError("Time samples must match in matrix_expect_val "
+                             "(use allow_times_differ=True to override)")
+
+    ##########
+
+    times = a.t
+    A = a
+    B = b
+
+    if (LM_clip is not None):
+        A = A[:,LM_clip]
+        B = B[:,LM_clip]
+
+    if (t_clip is not None):
+        times = t_clip
+        A = A.interpolate(t_clip)
+        B = B.interpolate(t_clip)
+
+    ##########
+    ## Time for actual math!
+
+    Abar_data  = np.conj(A.data)
+    B_data     = B.data
+
+    expect_val = np.zeros(times.size, dtype=np.complex)
+
+    ell_min = LM_clip.start
+    ell_max = LM_clip.stop - 1
+
+    for ellp, mp, ell, m, M_el in M(ell_min, ell_max):
+        i_a = LM_index(ellp, mp, A.ell_min)
+        i_b = LM_index(ell , m , B.ell_min)
+
+        expect_val += Abar_data[:, i_a] * M_el * B_data[:, i_b]
+
+    return times, expect_val
+
+def id_M(ell_min, ell_max):
+    """Generator for identity matrix elements (for use with matrix_expect_val)
+
+    Matrix elements yielded are
+    <ellp, mp|I|ell, m> = \delta_{ellp ell} \delta_{mp, m}
+    """
+
+    for ell in xrange(ell_min, ell_max+1):
+        for m in xrange(-ell, ell+1):
+            yield ell, m, ell, m, 1.
+
+def j_z_M(ell_min, ell_max):
+    """Generator for j^z matrix elements (for use with matrix_expect_val)
+
+    Matrix elements yielded are
+    <ellp, mp|j^z|ell, m> = 1.j * m \delta_{ellp ell} \delta_{mp, m}
+    This follows the convention for j^z from Ruiz+ (2008) [0707.4654]
+    """
+
+    for ell in xrange(ell_min, ell_max+1):
+        for m in xrange(-ell, ell+1):
+            yield ell, m, ell, m, (1.j * m)
+
+def make_j_plus_minus(sign):
+    """Produce the function j_plus_M or j_minus_M, based on sign.
+
+    The conventions for these matrix elements, to agree with Ruiz+
+    (2008) [0707.4654], should be:
+    <ellp, mp|j^+|ell, m> = 1.j * np.sqrt((l-m)(l+m+1)) \delta{ellp ell} \delta{mp (m+1)}
+    <ellp, mp|j^-|ell, m> = 1.j * np.sqrt((l+m)(l-m+1)) \delta{ellp ell} \delta{mp (m-1)}
+
+    The spinsfast function ladder_operator_coefficient gives
+    ladder_operator_coefficient(l,m) = np.sqrt((l-m)(l+m+1))
+
+    TODO
+    """
+
+    sign = float(sign)
+
+    if (sign != 1.) and (sign != -1.):
+        raise ValueError("sign must be either 1. or -1. in make_j_plus_minus")
+
+    def j_pm_M(ell_min, ell_max):
+        for ell in xrange(ell_min, ell_max+1):
+            for mp in xrange(-ell, ell+1):
+                m = mp + 1. * sign
+                if ((m < -ell) or (m > ell)):
+                    continue
+                yield ell, mp, ell, m, (1.j * ladder(ell, m*sign))
+
+    return j_pm_M
+
+j_plus_M  = make_j_plus_minus(1.)
+j_minus_M = make_j_plus_minus(-1.)
--- a/mode_calculations.py
+++ b/mode_calculations.py
@@ -898,17 +898,15 @@ def energy_flux(h):
         raise ValueError("Momentum flux can only be calculated from a `WaveformModes` object; "
                          +"this object is of type `{0}`.".format(type(h)))
     if h.dataType == hdottype:
-        hdot = h
+        hdot = h.data
     elif h.dataType == htype:
-        hdot = h.copy()
-        hdot.dataType = hdottype
-        hdot.data = h.data_dot
+        hdot = h.data_dot
     else:
         raise ValueError("Input argument is expected to have data of type `h` or `hdot`; "
                          +"this waveform data has type `{0}`".format(h.data_type_string))
 
     # No need to use matrix_expectation_value here
-    Edot = np.einsum('ij, ij -> i', hdot.data.conjugate(), hdot.data).real
+    Edot = np.einsum('ij, ij -> i', hdot.conjugate(), hdot).real
 
     Edot /= 16.*np.pi
 
--- a/mode_calculations.py
+++ b/mode_calculations.py
@@ -861,3 +861,50 @@ def _make_p_plus_minus_for_s_minus_2(sign):
 
 p_plus_M_for_s_minus_2  = _make_p_plus_minus_for_s_minus_2(1.)
 p_minus_M_for_s_minus_2 = _make_p_plus_minus_for_s_minus_2(-1.)
+
+def momentum_flux(h):
+    """Compute momentum flux from waveform
+    
+    This implements Eqs. (6)-(8) from Gerosa, Hébert, Stein <https://arxiv.org/abs/1802.04276>
+    """
+    from .waveform_modes import WaveformModes
+    from . import h as htype
+    from . import hdot as hdottype
+    if not isinstance(h, WaveformModes):
+        raise ValueError("Momentum flux can only be calculated from a `WaveformModes` object; "
+                         +"this object is of type `{0}`.".format(type(h)))
+    if h.dataType == hdottype:
+        hdot = h
+    elif h.dataType == htype:
+        hdot = h.copy()
+        hdot.dataType = hdottype
+        hdot.data = h.data_dot
+    else:
+        raise ValueError("Input argument is expected to have data of type `h` or `hdot`; "
+                         +"this waveform data has type `{0}`".format(h.data_type_string))
+    pdot = np.zeros((hdot.n_times, 3), dtype=float)
+    ell, m = hdot.LM.T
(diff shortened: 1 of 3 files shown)


duetosymmetry:
I have verified that this implementation of `momentum_flux` agrees with the one on branch [momentum_flux](https://github.com/moble/scri/tree/momentum_flux) using `np.allclose`. I do not have tests for the energy flux or angular momentum flux.

mode_calculations.py
```diff
@@ -600,3 +601,366 @@ def inner_product(t, abar, b, axis=None, apply_conjugate=False):
...
+
+    for ellp, mp, ell, m, M_el in M(ell_min, ell_max):
+        # QUESTION: Should we test if M_el is very close to 0, and
+        # continue to the next element if it's close enough?
+
+        i_a = LM_index(ellp, mp, A.ell_min)
+        i_b = LM_index(ell , m , B.ell_min)
+
+        # QUESTION: Do any python implementations short-circuit the
+        # multiplication if M_el is 0?
```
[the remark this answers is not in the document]
duetosymmetry: Alright, then it would be incumbent upon anybody who creates one of these sparse matrices to do their due diligence and only emit nonzero matrix elements.

mode_calculations.py
```diff
@@ -600,3 +601,366 @@ def inner_product(t, abar, b, axis=None, apply_conjugate=False):
...
+    B_data     = B.data
+
+    expect_val = np.zeros(times.size, dtype=np.complex)
+
+    ell_min = LM_clip.start
+    ell_max = LM_clip.stop - 1
+
+    for ellp, mp, ell, m, M_el in M(ell_min, ell_max):
+        # QUESTION: Should we test if M_el is very close to 0, and
+        # continue to the next element if it's close enough?
```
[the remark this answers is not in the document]
duetosymmetry: Point taken (but there is only one mode from `a` and one mode from `b` that multiply this matrix element; but importantly, we don't know any 'typical size' or if there are cancellations in the sum, so it's best avoid skipping terms in the sum)

duetosymmetry:
Indeed, you're right that my implementation is like `dok_matrix` (except there is no storage for the matrix — the elements are just generated one at a time).

I am trying to understand your suggestion. Is the following correct? Make a function which returns three arrays. Two arrays of indices `id_a`, `id_b`, and a third array of complex matrix elements `M_el`. The length of each of these is the number of nonvanishing elements. Then perform the sum `a[:,id_a] * M_el * b[:,id_b]` (with the broadcasting arranged correctly). Will that be more efficient than the generator-based implementation?

mode_calculations.py
```diff
@@ -600,3 +601,366 @@ def inner_product(t, abar, b, axis=None, apply_conjugate=False):
         integrand = np.conjugate(abar)*b
 
     return sdi(integrand, t, axis=axis)
+
+
+def matrix_expect_val(a, M, b,
```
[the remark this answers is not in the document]
duetosymmetry: Will rename.

mode_calculations.py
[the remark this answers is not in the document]
moble: I say numerous because there are usually tens of thousands of such elements, since `a` and `b` are functions of time, so "close to" might change in time.

moble:
Oh, I actually think that you should use `dia_matrix` since the infrastructure is already there (and scipy is actually a requirement for this package anyway).  There are lots of ways to convert/construct scipy's sparse matrices, but I'd think that the cleanest one here is to construct the bands directly.  For example, `j_z_M` would just be
```python
band = [1j*m for ell in range(ell_min, ell_max+1) for m in range(-ell, ell+1)]
return dia_matrix(([band], [0]), shape=(n_modes, n_modes))
```
and `j_pm_M` would be something like
```python
offset = sign
band = [1.j * ladder(ell, (mp+sign)*sign) if abs(mp+sign)<=ell else 0
        for ell in range(ell_min, ell_max+1) for mp in range(-ell, ell+1)]
return dia_matrix(([band], [offset]), shape=(n_modes, n_modes))
```
(though I haven't checked this in detail).  I wouldn't really sweat the efficiency of the construction of these too much, since they're probably tiny compared to the other operations, but they could be jitted as well.

duetosymmetry:
Some operators will be dense in diagonal storage, but I think some will have a lot of zeros. For example, here's what the indices of non-vanishing matrix elements as pairs `(id_a, id_b)` are from `p_plus_M_for_s_minus_2(ell_min=2, ell_max=6)`, sorted so that diagonals appear together:
```
[(21, 32), (22, 33), (23, 34), (24, 35), (25, 36), (26, 37), (27, 38), (28, 39), (29, 40), (30, 41), (31, 42),
 (12, 21), (13, 22), (14, 23), (15, 24), (16, 25), (17, 26), (18, 27), (19, 28), (20, 29),
 (5, 12), (6, 13), (7, 14), (8, 15), (9, 16), (10, 17), (11, 18),
 (0, 5), (1, 6), (2, 7), (3, 8), (4, 9),
 (1, 0), (2, 1), (3, 2), (4, 3), (6, 5), (7, 6), (8, 7), (9, 8), (10, 9), (11, 10), (13, 12), (14, 13), (15, 14), (16, 15), (17, 16), (18, 17), (19, 18), (20, 19), (22, 21), (23, 22), (24, 23), (25, 24), (26, 25), (27, 26), (28, 27), (29, 28), (30, 29), (31, 30), (33, 32), (34, 33), (35, 34), (36, 35), (37, 36), (38, 37), (39, 38), (40, 39), (41, 40), (42, 41), (43, 42), (44, 43),
 (7, 0), (8, 1), (9, 2), (10, 3), (11, 4),
 (14, 5), (15, 6), (16, 7), (17, 8), (18, 9), (19, 10), (20, 11), (23, 12), (24, 13), (25, 14), (26, 15), (27, 16), (28, 17), (29, 18), (30, 19), (31, 20),
 (34, 21), (35, 22), (36, 23), (37, 24), (38, 25), (39, 26), (40, 27), (41, 28), (42, 29), (43, 30), (44, 31)]
```
There are a lot of zeros in these diagonals. So is it more efficient to build a `dia_matrix`, or to handle indexing lists `id_a, id_b` and an array of matrix elements?

duetosymmetry:
I just did a little bit of profiling, here's what I found.

Most of the time in `energy_flux`, `momentum_flux`, and `angular_momentum_flux` is spent in performing the spline interpolation for the time derivative. I used a synthetic (QNM) waveform with l=2..8, with 5001 time samples. Performing `w.momentum_flux()` took ~144ms, but performing `wdot.momentum_flux()` took ~40ms.

So, lesson 1: pre-compute the derivative and use that when possible. [Question: can we make `WaveformBase.data_dot()` be memoizing? If nobody uses it, no gain or penalty. If somebody saves it, then it should be the same object reference, so again no gain or penalty. If the user treats it as a temporary, then memoizing is beneficial. As an aside, I think there should be a time derivative operator that returns a new Waveform object, not just the data].

I also tried to speed things up by rewriting the inner part of `matrix_expect_val` (whose name will change) to do something like this instead:
```
a_inds = []
b_inds = []
M_els  = []

for ellp, mp, ell, m, M_el in scri.mode_calculations.p_plus_M_for_s_minus_2(ell_min, ell_max):
    a_inds.append(LM_index(ellp, mp, wm.ell_min))
    b_inds.append(LM_index(ell , m , wm.ell_min))
    M_els.append(M_el)

expect_val = Abar_data[:, a_inds] * B_data[:, b_inds] @ M_els
```
Then on this same test waveform, instead of ~40ms it took ~50ms. I imagine that most of that penalty is because there are temporaries created for `Abar_data[:, a_inds]`, `B_data[:, b_inds]`, and the intermediate product, which are allocated/deallocated. So, lesson 2: I have not found a way to be fancier indexing or sparse matrix approach to get a speedup.

moble:
> Some operators will be dense in diagonal storage, but I think some will have a lot of zeros.

I forgot about coupling between neighboring ell values.  I still see a factor of ~2 speedup over your original code when using `dia_matrix` for this pattern.  That said, your approach becomes ~2 faster than mine (for all the matrix patterns) once I use numba.  So I suggest something like this:
```python
@jit
def sparse_expectation_value(a, rows, columns, values, b):
    n_times = a.shape[0]
    n_elements = rows.shape[0]
    expectation_value = np.zeros(n_times, dtype=np.complex_)
    for i_time in range(n_times):
        for i_element in range(n_elements):
            expectation_value[i_time] += (
                a[i_time, rows[i_element]]
                * values[i_element]
                * b[i_time, columns[i_element]]
            )
    return expectation_value
```


> lesson 1: pre-compute the derivative and use that when possible

I agree.  On a related note, I was gonna suggest reorganizing the top-level functions to make the code assume that you want all fluxes, so that they could be computed more efficiently all at once...

> can we make WaveformBase.data_dot() be memoizing?

I go back and forth on this.  In the past, I've come down on the side of not doing this because it's so easy to save and re-use the output when needed, and it's just so much data that it feels leaky to keep it around automatically.  Of course, this is informed by how I use it, and it has actually been convenient to save and re-use for me.  The only real consequence is that functions would need a separate `hdot` parameter (or similar), which isn't as cute and OOPy, but feels like a tiny and reasonable inconvenience.  So, for example, your flux functions could have `hdot=None` arguments that just compute it when needed.  I don't have very strong feelings about it.

> As an aside, I think there should be a time derivative operator that returns a new Waveform object, not just the data

My main worry about this is that it can break the promise that a Waveform object actually knows what it contains.  (Dante and I had a similar conversation about eth the other day.)  As you know, there is an `hdot` type, but nothing other than `h` knows what it becomes when differentiated.  So I guess we could do the right thing for `h`, and just make everything else `UnknownDataType`, but I'm also really not sure of what the benefit would be.

duetosymmetry:
Aha, I should have profiled in greater depth! I can see why the code you wrote would play well with numba. I will try to restructure things into the pattern you suggested. The only thing I dislike about it is that it would make the matrix element functions (e.g. `p_z_M...`) have to emit array-indices rather than SWSH-indices, making those functions lower level. That's probably fine, though, since they already must accept `ell_min` and `ell_max` as arguments. It just means that every one of those functions will have repeated code of `a_inds.append(LM_index(ellp, mp, wm.ell_min))` and so on. However I'm already violating DRY, because already each of the matrix element functions has 2, 3, or 4 nested for loops in the exact same structure. I don't know how to avoid repeating myself in this case.

> On a related note, I was gonna suggest reorganizing the top-level functions to make the code assume that you want all fluxes, so that they could be computed more efficiently all at once

Totally agree, I came to the same conclusion.

Re: memoizing: I though better of this suggestion... Right now, the user is free to edit the data within a WaveformBase object. If for some reason the user computes the time derivative, then re-uses the storage and wants to recompute the time derivative, they would need to manually invalidate the cache. This sounds like a headache. Better to avoid caching if the object's data is writeable.

> the promise that a Waveform object actually knows what it contains. (Dante and I had a similar conversation about eth the other day.) As you know, there is an hdot type, but nothing other than h knows what it becomes when differentiated.

The only way I know how to address this idea is to create algebraic data types on top of python. That is possible but does sounds like a pain.

PS Do you have recommendations for names for the matrix-element-generating functions? They are not very nice names right now.

mode_calculations.py
```diff
@@ -600,3 +601,366 @@ def inner_product(t, abar, b, axis=None, apply_conjugate=False):
...
+
+def j_z_M(ell_min, ell_max):
+    """Generator for j^z matrix elements (for use with matrix_expect_val)
+
+    Matrix elements yielded are
+    <ellp, mp|j^z|ell, m> = 1.j * m \delta_{ellp ell} \delta_{mp, m}
+    This follows the convention for j^z from Ruiz+ (2008) [0707.4654]
+    """
+
+    for ell in xrange(ell_min, ell_max+1):
```
moble: FYI, `xrange` is a holdover from python 2.  Since `scri` is now evidently just a python 3 package, you should feel free to just use `range`.

mode_calculations.py
```diff
@@ -600,3 +601,366 @@ def inner_product(t, abar, b, axis=None, apply_conjugate=False):
...
+        hdot = h
+    elif h.dataType == htype:
+        hdot = h.copy()
+        hdot.dataType = hdottype
+        hdot.data = h.data_dot
+    else:
+        raise ValueError("Input argument is expected to have data of type `h` or `hdot`; "
+                         +"this waveform data has type `{0}`".format(h.data_type_string))
+
+    _, Edot  = matrix_expect_val( hdot, id_M,  hdot )
```
moble: It might be worthwhile using something simpler here.  For example, `Edot = np.einsum('ij, ij -> i', hdot.conjugate(), hdot)` is ~4 times faster, even without the rest of the overhead.

moble:
> The only thing I dislike about it is that it would make the matrix element functions (e.g. p_z_M...) have to emit array-indices rather than SWSH-indices, making those functions lower level.

The function I wrote just above should only form the core computation, and you should still keep all the other logic about clipping and so on.  So there's no reason to repeat yourself with all the array-indexing logic; you could just keep the conversion inside that bigger function.

> Do you have recommendations for names for the matrix-element-generating functions? They are not very nice names right now.

Among other wisdom provided by `import this`, we find "Namespaces are one honking great idea -- let's do more of those!"  Since these functions are inside a submodule, I have no problem  with very short names like `j_z`, `j_plus`, etc.  Of course, the ultimate namespace is inside a function, and I can't think of any other uses for these matrices, so I would find it reasonable to actually keep those functions within their respective flux functions — `j_z` fully defined inside `angular_momentum_flux`, and so on.

On a related note, how would you feel about moving all of this to a `flux` submodule of its own?

mode_calculations.py
duetosymmetry: Done

mode_calculations.py
duetosymmetry: Done (for the code I added, I did not touch instances of `xrange` in other places)

duetosymmetry:
I've added another commit (this whole PR may need to be squashed down when it's ready, but right now I wanted these changes to be easy to see).

I put in the jitted inner loop, which is very helpful. Even more helpful was adding a `poincare_fluxes` function which returns edot, pdot, jdot and only has to calculate a time derivative once (or 0 times, if it is passed into the optional argument).

I have not yet migrated helper functions inside of driver functions, because of my lack of knowledge of python's inner workings: If I define a function `inner()` inside the scope of another function `outer()`, the next time I call `outer()`, does the python interpreter have to re-evaluate the definition? Or does it detect if any variables were captured from an outside scope, and if not, avoid re-parsing? Or is this all moot, because the default python interpreter actually doesn't do anything clever, so there is no penalty to a local function definition, and it only matters with jitted functions?

Regarding moving the functions into another module: This may be reasonable. I think the name `flux` is too restrictive (though I don't have any examples of things to compute with these tools that are not considered a flux). Would you refactor any of the functions `LdtVector(W)`, `LVector(W1,W2)` etc. to use `matrix_expectation_value()`? It may give a small speed penalty, but these functions are a bit hard to read right now. If they are refactored then they would have to import from the other module.

duetosymmetry:
PS I think `LdtVector` should be related to the flux of angular momentum in some simple way, but I am not getting agreement between the two. The z components agree (except for a factor fo 1/(16\pi), which is expected). But the x and y components do not agree.

mode_calculations.py
```diff
@@ -600,3 +601,422 @@ def inner_product(t, abar, b, axis=None, apply_conjugate=False):
...
+    from . import hdot as hdottype
+    if not isinstance(h, WaveformModes):
+        raise ValueError("Momentum flux can only be calculated from a `WaveformModes` object; "
+                         +"this object is of type `{0}`.".format(type(h)))
+    if h.dataType == hdottype:
+        hdot = h
+    elif h.dataType == htype:
+        hdot = h.copy()
+        hdot.dataType = hdottype
+        hdot.data = h.data_dot
```
10220: Since you are just passing in arrays as arguments to `np.einsum` on line 911, I don't think making the new WaveformModes objects here are necessary. You could just do something like:
```
if h.dataType = hdottype:
    hdot = h.data
elif h.dataType == htype:
    hdot = h.data_dot
```
Then in your `np.einsum` command, you just have:
```
Edot = np.einsum('ij, ij -> i', hdot.conjugate(), hdot).real
```

moble:
> If I define a function `inner()` inside the scope of another function `outer()`, the next time I call `outer()`, does the python interpreter have to re-evaluate the definition? Or does it detect if any variables were captured from an outside scope, and if not, avoid re-parsing?

When you first *load* `outer`, python compiles it and its contents to bytecode.  So as part of this process it compiles `inner` to bytecode, and as part of that process it figures out which scope the variables come from.  (E.g., if `inner` is a closure, that is recorded in the bytecode.)  The resulting bytecode has one "code object" for `inner` and another one for `outer`.  So in that sense, python does not have to re-evaluate or re-parse anything; it's all done with understanding the code.  (If you want to geek out on this, look into the standard-library `dis` module.)

However, it is true that when you actually *call* `outer`, part of that process involves creating a "function object" for `inner`.  It's sort of like the "code object" is a class, and the "function object" is instantiated from that class.  But this basically amounts to creating a struct and filling in the values — and in particular, copying any values that need to come from another scope — as opposed to just filling in the values if `inner` had been defined outside of `outer`.  So there is actually some overhead to doing things in this way.  But it's probably of order 1 microsecond even with your complicated functions.

---

I was about to start in on some decorator fanciness to make those nicer still.  But at this point, the savings probably don't matter that much, and anything more would mostly just be satisfying my own propensity for geeking out.  I'm also guessing you don't have time for this and just want me to merge your already excellent contributions.  I would be happy to do so as soon as you incorporate Dante's fix and/or move to the new file, if you want.  Maybe I'll add the niceties myself after we've got tests for these functions.  I should probably refactor `LdtVector`, etc., at some point but I have no qualms about importing from another submodule when the time comes.

In short, this looks mergeable to me, so just let me know when you're happy with it.

mode_calculations.py
duetosymmetry: Good point! Done, will push a commit in a min.

duetosymmetry:
I've incorporated the suggestion by @10220, and also moved the helper functions inside of the user-facing functions. This code is still inside of mode_calculations.py for the time being.

I also did some comparisons against `LdtVector`. I find that:
```
h.LdtVector()/16./np.pi + h.angular_momentum_flux()
```
is approximately zero. Specifically, if I take the fractional error of each component of the vector difference, and then take the vector norm, and look at the timeseries of (norm of fractional-error-vectors as func of time), this norm is typically 2e-15 [for a precessing q=2 system over 5000M], but sometimes goes up to 2e-13 (I do not see anything special at that time). In terms of a time comparison, `h.angular_momentum_flux()` for this waveform takes 388 ms ± 22.6 ms . If I precompute `hdot`, then `h.angular_momentum_flux(hdot)` takes 64 ms ± 533 µs. `h.LdtVector()` always computes the spline derivative so it takes 330 ms ± 8.46 ms.

duetosymmetry:
PS I think this is mergeable, please let me know if there's anything else you'd like to see (e.g. squashing all the commits)

moble:
Thanks very much for all this, Leo!

(P.S. Github lets me squash and merge with one button, so that part's no problem.)

fmahebert:
@moble Will this update make its way to Pypi?

moble:
@fmahebert Yep.  I'm just working on some tests before I do that.

fmahebert:
@moble Okay, thanks! 🙂

moble:
@fmahebert This is now on pypi, and — far more importantly — on conda-forge.

PR #18, opened 2019-08-13
Convenient operators

10220:
This adds the ability to perform a quick second "time" derivative or eth/ethbar derivatives. One can also apply mulitply eths and combinations of eth & ethbars. 

**WARNING**: This currently throws a Numba deprecation warning from `spherical_functions`,
```
/home/dante/Documents/packages/miniconda3/lib/python3.6/site-packages/spherical_functions/mode_conversions.py:110: NumbaWarning: 
Compilation is falling back to object mode WITH looplifting enabled because Function "ethbar_GHP" failed type inference due to: Invalid use of Function(<built-in function array>) with argument(s) of type(s): (array(complex128, 2d, C))
 * parameterized
In definition 0:
    TypingError: array(complex128, 2d, C) not allowed in a homogeneous sequence
    raised from /home/dante/Documents/packages/miniconda3/lib/python3.6/site-packages/numba/typing/npydecl.py:460
In definition 1:
    TypingError: array(complex128, 2d, C) not allowed in a homogeneous sequence
    raised from /home/dante/Documents/packages/miniconda3/lib/python3.6/site-packages/numba/typing/npydecl.py:460
This error is usually caused by passing an argument of a type that is unsupported by the named function.
[1] During: resolving callee type: Function(<built-in function array>)
[2] During: typing of call at /home/dante/Documents/packages/miniconda3/lib/python3.6/site-packages/spherical_functions/mode_conversions.py (135)


File "../../../packages/miniconda3/lib/python3.6/site-packages/spherical_functions/mode_conversions.py", line 135:
def ethbar_GHP(modes, spin_weight, ell_min=0):
    <source elided>
    ell_max = int(sqrt(len(modes) + LM_total_size(0, ell_min - 1))) - 1
    ethbar_modes = np.array(modes)
    ^

```
This warning gets thrown for each of the eth functions. This error goes away if we remove the np.array() cast in the last line in the `spherical_functions` code but then things will break if someone passes in something that's not a numpy array. I played around a bit with getting this to work with if statements and try blocks but any attempt to acccess information about the array will throw the Numba error.

Any ideas?

### DIFF
--- a/waveform_modes.py
+++ b/waveform_modes.py
@@ -326,7 +326,6 @@ class WaveformModes(WaveformBase):
             Number of times to apply the spin lowering operator
         eth_convention: either 'NP' or 'GHP' 
             Choose between Newman-Penrose or GHP convention
-                Choose between Newman-Penrose or GHP convention.
 
         Returns
         -------
--- a/waveform_modes.py
+++ b/waveform_modes.py
@@ -339,6 +339,8 @@ class WaveformModes(WaveformBase):
         elif eth_convention == 'GHP':
             operator = sf.eth_GHP
             antioperator = sf.ethbar_GHP
+        else:
+            raise ValueError("eth_convention must either be 'NP' or 'GHP'; got '{}'".format(eth_convention))
         s = self.spin_weight
         mode_data = self.data.transpose()
         for n_eth in range(N_eth):
--- a/waveform_modes.py
+++ b/waveform_modes.py
@@ -315,15 +315,18 @@ class WaveformModes(WaveformBase):
             raise ValueError("Input `ell_m` should be an Nx2 sequence of integers")
         return ell_m[:, 0] * (ell_m[:, 0] + 1) - self.ell_min ** 2 + ell_m[:, 1]
 
-    def apply_eth(self, N_eth=0, N_ethbar=0, eth_convention='NP'):
-        """Apply spin raising/lowering operators to waveform mode data.
+    def apply_eth(self, operations, eth_convention='NP'):
+        """Apply spin raising/lowering operators to waveform mode data 
+           in a specified order. 
 
         Parameters
         ----------
-        N_eth: int
-            Number of times to apply the spin raising operator
-        N_ethbar: int
-            Number of times to apply the spin lowering operator
+        operations: str of combinations of '+' and/or '-'
+            The order of eth ('+') and ethbar ('-') operations to perform, 
+            applied from right to left. Example, operations='--+' will perform
+            on WaveformModes data f:
+                ethbar(ethbar(eth(f)))
+
         eth_convention: either 'NP' or 'GHP' 
             Choose between Newman-Penrose or GHP convention
 
@@ -334,34 +337,37 @@ class WaveformModes(WaveformBase):
         """
         import spherical_functions as sf
         if eth_convention == 'NP':
-            operator = sf.eth_NP
-            antioperator = sf.ethbar_NP
+            eth = sf.eth_NP
+            ethbar = sf.ethbar_NP
         elif eth_convention == 'GHP':
-            operator = sf.eth_GHP
-            antioperator = sf.ethbar_GHP
+            eth = sf.eth_GHP
+            ethbar = sf.ethbar_GHP
         else:
             raise ValueError("eth_convention must either be 'NP' or 'GHP'; got '{}'".format(eth_convention))
         s = self.spin_weight
         mode_data = self.data.transpose()
-        for n_eth in range(N_eth):
-            mode_data = operator(mode_data, s, self.ell_min)
-            s += 1
-        for n_ethbar in range(N_ethbar):
-            mode_data = antioperator(mode_data, s, self.ell_min)
-            s -= 1
+        if not set(operations).issubset({'+','-'}):
+            raise ValueError("Operations must be combinations of '+' and '-'; got '{}'".format(operations))
+        for operation in reversed(operations):
+            if operation == '+':
+                mode_data = eth(mode_data, s, self.ell_min)
+                s += 1
+            elif operation == '-':
+                mode_data = ethbar(mode_data, s, self.ell_min)
+                s -= 1
         return mode_data.transpose()
 
     @property
     def eth(self):
         """Returns the spin-raised waveform mode data.
         """
-        return self.apply_eth(N_eth=1)
+        return self.apply_eth(operations='+')
     
     @property
     def ethbar(self):
         """Returns the spin-lowered waveform mode data.
         """
-        return self.apply_eth(N_ethbar=1)
+        return self.apply_eth(operations='-')
 
     def inner_product(self, b,
                       t1=None, t2=None,
--- a/waveform_modes.py
+++ b/waveform_modes.py
@@ -316,7 +316,23 @@ class WaveformModes(WaveformBase):
         return ell_m[:, 0] * (ell_m[:, 0] + 1) - self.ell_min ** 2 + ell_m[:, 1]
 
     def apply_eth(self, N_eth=0, N_ethbar=0, eth_convention='NP'):
+        """Apply spin raising/lowering operators to waveform mode data.
 
+        Parameters
+        ----------
+        N_eth: int
+            Number of times to apply the spin raising operator
+        N_ethbar: int
+            Number of times to apply the spin lowering operator
+        eth_convention: either 'NP' or 'GHP' 
+            Choose between Newman-Penrose or GHP convention
+                Choose between Newman-Penrose or GHP convention.
+
+        Returns
+        -------
+        mode_data: array of complex 
+
+        """
         import spherical_functions as sf
         if eth_convention == 'NP':
             operator = sf.eth_NP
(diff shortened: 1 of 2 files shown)


waveform_modes.py
```diff
@@ -315,6 +315,50 @@ def indices(self, ell_m):
...
+        mode_data = self.data.transpose()
+        for n_eth in range(N_eth):
+            mode_data = operator(mode_data, s, self.ell_min)
+            s += 1
+        for n_ethbar in range(N_ethbar):
+            mode_data = antioperator(mode_data, s, self.ell_min)
+            s -= 1
+        return mode_data.transpose()
+
+    def eth(self):
```
moble: Add `@property` above this function and `ethbar`.

waveform_modes.py
```diff
@@ -315,6 +315,50 @@ def indices(self, ell_m):
...
+
+        Parameters
+        ----------
+        N_eth: int
+            Number of times to apply the spin raising operator
+        N_ethbar: int
+            Number of times to apply the spin lowering operator
+        eth_convention: either 'NP' or 'GHP' 
+            Choose between Newman-Penrose or GHP convention
+                Choose between Newman-Penrose or GHP convention.
```
moble: Repetition?

waveform_modes.py
```diff
@@ -315,6 +315,50 @@ def indices(self, ell_m):
...
+
+        """
+        import spherical_functions as sf
+        if eth_convention == 'NP':
+            operator = sf.eth_NP
+            antioperator = sf.ethbar_NP
+        elif eth_convention == 'GHP':
+            operator = sf.eth_GHP
+            antioperator = sf.ethbar_GHP
+        s = self.spin_weight
```
moble: Add an `else` clause that raises a ValueError and explains that it has to be either NP or GHP.

waveform_modes.py
```diff
@@ -315,6 +315,50 @@ def indices(self, ell_m):
             raise ValueError("Input `ell_m` should be an Nx2 sequence of integers")
         return ell_m[:, 0] * (ell_m[:, 0] + 1) - self.ell_min ** 2 + ell_m[:, 1]
 
+    def apply_eth(self, N_eth=0, N_ethbar=0, eth_convention='NP'):
```
moble: Since this function signature enables it, and it is sometimes useful to consecutively apply various operators, I wonder if it would be better to use some way of specifying the order of operators.  For example, I'm sure I've seen eth(eth(ethbar(ethbar(f)))) [or the conjugate], and I'm pretty sure I've sure eth(ethbar(eth(ethbar(f)))).  So maybe instead of `N_eth` and `N_ethbar`, there could be an argument like `operations` that takes a string of `'+'` for raise and `'-'` for lower.  Then, my first example would be `'++--'` and my second would be `'+-+-'`.  What do you think?

10220: Ahh that's a great idea! that'll be much for flexible, I'll implement this.

moble: I just noticed a little ambiguity here regarding the order of operations.  Obviously, eth should be '+' and ethbar should be '-'.  I think what I wrote is the most appropriate choice, so for example eth(ethbar(f)) would be written as '+-'.  But note that this means applying ethbar *first*, which means iterating through the string in *reverse* order:
```python
for operation in reversed(operations):
```
Of course, people might expect the opposite order, so this should be emphasized in the docstring.

It's also probably best to check for valid operations before starting that loop.  Here's a simple way to do it:
```python
if not set(operations).issubset({'+', '-'}):
    raise ValueError("Operations can be combinations of '+' and '-'; got '{0}'".format(operations)
```

moble:
Regarding the numba warning,

> This error goes away if we remove the np.array() cast in the last line in the spherical_functions code but then things will break if someone passes in something that's not a numpy array.

It's worse than that, because that `np.array` call not only casts, but also copies.  If you remove it, the function will actually be modifying the array that was passed in, which we definitely don't want.

It turns out that this is a bug in numba, for which I've opened [an issue](https://github.com/numba/numba/issues/4470).  I think it's reasonable to just ignore the warning until it's fixed in numba.

10220:
Yikes, I was afraid of that. I can open a PR in spherical_functions to have `@jit(warn=False)` for the `eth_NP` and `eth_GHP` functions until numba addresses the issue?

moble:
No, I think it's really not worth the trouble, and if we forget to change it back once they've fixed it, we might miss future problems.  If you really hate seeing the warnings, you can [filter them](https://numba.pydata.org/numba-doc/dev/reference/deprecation.html#suppressing-deprecation-warnings) in whatever code you're using to call this code.  But it's generally bad policy to alter libraries to deal with temporary upstream warnings.

moble:
Thanks!

duetosymmetry:
Hi @10220 and @moble,
I'm happy to see eth and ethbar make it in but have a minor suggestion about performance. Currently, the `apply_eth` operation will perform k multiplications on each (l,m) mode, where k is the length of the string of operations to perform. There will also be k*l square roots taken (but this is a minor cost compared to the first issue). These can both be solved by looping over l, and computing the combined ladder factors of the sequence of eth's and ethbar's for each l, which only requires one square root; and then only performing one multiple for each (l,m) mode. For example, the ladder factors can be computed like this:
```python
def ladder_factor(operations, s, ell, eth_conventions='NP'):
    op_dict = {u'ð': +1, u'ð̅': -1, '+': +1, '-': -1, +1:+1, -1:-1}
    conv_dict = {'NP': 1., 'GHP': 0.5}

    convention_factor = conv_dict[eth_conventions]

    ladder = 1.
    prefactor = 1.

    _s = float(s)

    for op in reversed(operations):
        sign = op_dict[op]
        prefactor *= sign
        ladder *= (ell - _s * sign) * (ell + _s * sign + 1.)
        ladder *= convention_factor
        _s += sign
    
    ladder = np.sqrt(ladder) * prefactor
    return ladder
```

Two asides: (i) using a dict for `op_dict` and `conv_dict` means that unknown symbols in `operations` or unknown convention names in `eth_conventions` will result in raising a `KeyError`, which seems like an appropriate thing to do; (ii) using a dict for `op_dict` gives multiple ways of specifying the list of operations, e.g.:
```python
ladder_factor('++-', -2, 5, eth_conventions='NP') # current approach
ladder_factor([+1,+1,-1], -2, 5, eth_conventions='NP') # valid!
ladder_factor(['ð','ð','ð̅'], -2, 5, eth_conventions='NP') # also valid!
```
Unfortunately 'ððð̅' doesn't work because ð̅ uses a combining unicode character, and I haven't figured out how to get python to split unicode strings while keeping combining characters attached too what they modify. I wouldn't really want to allow ð̅ as a key, because of this potential pitfall... just wanted to raise the possibility though.

moble:
Good points, Leo.

One pitfall to watch out for is that `ell` could be less than `abs(s)` on input, so it's safer and simpler to check for that here and return 0, rather than hoping that it's avoided in the calling code.  (I'm also not sure why you use `_s` instead of just `s`, since floats are immutable... or why you cast it to a float...)

FWIW, this ugmo is the nicest way I can think of (that doesn't require external dependencies) to split the string:
```python
from unicodedata import combining
iterator = reversed(operations)
reversed_operations = [c if not combining(c) else next(iterator)+c for c in iterator]
```
But it would probably be easier to just use
```python
operations.replace('ð̅', '-').replace('ð', '+')
```
It would also make a lot of sense to me to just normalize `operations` to a list of `1`s and `-1`s, which would allow strings, lists of strings, or lists of ints, and would obviate the need for a sign lookup since the operation would *be* the sign.

I actually strongly dislike raising KeyError for things like this, because every time I get such an error from some unfamiliar piece of code, it means I have to go in and understand what that code is doing, why it's using a dict that I've never seen before, how that relates to my input, and whether the failure is some bug in their code or my error on input values.  Returning a descriptive ValueError is much clearer, in my experience.

duetosymmetry:
> One pitfall to watch out for is that ell could be less than abs(s) on input, so it's safer and simpler to check for that here and return 0, rather than hoping that it's avoided in the calling code. (I'm also not sure why you use _s instead of just s, since floats are immutable... or why you cast it to a float...)

I imagined this as a local helper function, but if this was a publicly available function, I agree that it should do input validation.

Yes, I was worried about mutability of s – I haven't yet internalized what gets passed by value or reference.

I think this should cover input validation and the case of ell < abs(s):

```python
def ladder_factor(operations, s, ell, eth_conventions='NP'):
    op_dict = {u'ð': +1, u'ð̅': -1, '+': +1, '-': -1, +1: +1, -1: -1}
    conv_dict = {'NP': 1., 'GHP': 0.5}

    # Normalize combining unicode characters
    if (isinstance(operations, str)):
        operations = operations.replace('ð̅', '-').replace('ð', '+')
    
    if not set(operations).issubset(op_dict.keys()):
        raise ValueError("operations must be a string composed of "
                         "'ð', 'ð̅', '+', '-', or a list with "
                         "elements coming from the set {}"
                         .format(set(op_dict.keys())))
    
    if eth_conventions not in conv_dict:
        raise ValueError("eth_conventions must be one of {}"
                         .format(set(conv_dict.keys())))

    convention_factor = conv_dict[eth_conventions]

    ladder = 1.
    sign_factor = 1.

    for op in reversed(operations):
        sign = op_dict[op]
        sign_factor *= sign
        ladder *= (ell - s * sign) * (ell + s * sign + 1.) if (ell >= abs(s)) else 0.
        ladder *= convention_factor
        s += sign
    
    ladder = sign_factor * np.sqrt(ladder)
    return ladder
```

Now this is possible:
```python
ladder_factor('ðð̅', -2, 4)
ladder_factor('ðð̅+-', -2, 4) # Mixing string conventions is allowed
ladder_factor(['ð̅', -1, '+'], -2, 3) # Seems ridiculous but also technically valid
```
Not that I recommend writing code like this...

PR #19, opened 2019-08-13
Fix plotting bug

10220:
This fixes the bug when trying to make plots with the COM correction.


SpEC/com_motion.py
```diff
@@ -142,7 +142,6 @@ def estimate_avg_com_motion(path_to_horizons_h5='Horizons.h5',
         max_d_color = min(1.0, 10*max_displacement)
         fig = plt.figure(figsize=(10, 7))
         plt.plot(t, com, alpha=0.25, lw=1.5)
-        plt.gca().set_color_cycle(None)
```
moble: Don't remove this; just replace `set_color_cycle` with the newer function: `set_prop_cycle`.  It's important to keep the corresponding line properties for the next line.

moble:
Thanks!

PR #20, opened 2019-08-13
Support supertranslation for all Weyl scalars

10220:
Implements Eq. (17) of  [_Transformations of asymptotic gravitational-wave data_](https://arxiv.org/abs/1509.00862v3) to provide supertranslations to all Weyl scalars.

PR #21, opened 2019-08-22
Support supertranslation of all weyl scalars

10220:
Implements Eq. (17) of  [_Transformations of asymptotic gravitational-wave data_](https://arxiv.org/abs/1509.00862v3) to provide supertranslations to all Weyl scalars. This reopens PR #20 after testing.

### DIFF
--- a/waveform_grid.py
+++ b/waveform_grid.py
@@ -350,6 +350,18 @@ class WaveformGrid(WaveformBase):
         boost_velocity : float array of length 3, optional
             This is the three-velocity vector of the grid frame relative to the mode frame.  The norm of this vector
             is checked to ensure that it is smaller than 1.  Defaults to [], corresponding to no boost.
+        psi4_modes : WaveformModes
+            The object storing the modes of a Psi4 waveform. This is only used to compute a supertranslation of a psi3,
+            psi2, psi1, or psi0 waveform. 
+        psi3_modes : WaveformModes
+            The object storing the modes of a Psi4 waveform. This is only used to compute a supertranslation of a psi2, 
+            psi1, or psi0 waveform. 
+        psi2_modes : WaveformModes
+            The object storing the modes of a Psi4 waveform. This is only used to compute a supertranslation of a psi1 
+            or psi0 waveform. 
+        psi1_modes : WaveformModes
+            The object storing the modes of a Psi4 waveform. This is only used to compute a supertranslation of a psi0 
+            waveform. 
 
         Returns
         -------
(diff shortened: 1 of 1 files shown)


10220:
Surprisingly, constructing the Wigner D matrices (i.e. calls to `spherical_functions.SWSH_grid`) only takes up 1% of the computational expense in performing the supertranslation even for Psi0. The computational expense is dominated by interpolation.

moble:
Oh, of course.  I knew that.

But the same logic applies: to transform Psi0, Psi1, Psi2, Psi3, and Psi4, we need to interpolate Psi4, so why do that separately for each transformation?  It'll still be more efficient to do them all at once.

10220:
That's very true. I'll make a function like `transform_Bondi_system` that will apply the transformation to all of the supplied quantities at once so everything can be stored. However, it might still be nice to have the option to do the transformation as a class method. For example, all I need for my own purposes is to transform h and Psi2 which would not benefit from this optimization.

moble:
Oh, I see.  So you transform first, and then interpolate.  Looking back at my `from_modes` function, it looks like this is also how I did it.  However, that was for a simpler application: just transforming a single waveform.  Because we'll need something fancier, I was just assuming that you would do things in the opposite order, without thinking back to my original code.  So I should probably explain why I expect that the opposite order will be better.

First, because you've found that interpolation is the slowest part, and because construction of the interpolators is significantly more expensive than evaluating them (typically by a factor of 3 or 4), it would be nice to just construct them once, store them, and then evaluate as needed.  This means that you can't implement this as just a function; it should be a simple class to store these things.  So the way I see this working is that you'd convert each WaveformModes object to a WaveformGrid, then just store a list of `InterpolatedUnivariateSpline`s for each grid point.  All that is a one-time cost.  Then, as needed, you would interpolate each grid point to the appropriate set of times.  You'd have to convert to modes, evaluate at the new grid points, combine the values appropriately, and transform back to modes again.  There are a couple extra steps in there, but there's room for improvement in the code that does them, and it sounds like those would still be less costly than the interpolation.

Second, we need to keep in mind the objective of generalizing the nice-section equation to a condition that can be applied over some span of time, rather than on a single slice.  Assuming that this takes the form of minimizing some L2-norm measure, it'll probably be extremely helpful to implement the derivative of that measure (with respect to the parameters of the BMS transformation).  I believe that derivative will require values of the *untransformed* fields on the new time slices and in the new spherical coordinates, which are obtained as part of the process outlined above.  Obviously, this is an item for future work, but it's important to anticipate it so that you don't have to start from scratch on the code once the time comes.

PR #22, opened 2019-10-10
Speed up WaveformModes.apply_eth: only one multiply per element

duetosymmetry:
This follows the discussion starting in this PR comment: https://github.com/moble/scri/pull/18#issuecomment-523710834

Also only one square root per value of \ell, but this is relatively minor compared to only one multiply per element.


duetosymmetry:
@10220 @moble would either of you like to review this PR?

duetosymmetry:
Rebased onto upstream

duetosymmetry:
Rebased so this PR is a fast-forward

moble:
Thanks.  Sorry for the slow reply.  :$

PR #23, opened 2019-10-10
Add properties for first and second time integrals.

duetosymmetry:
Not stopping user from shooting themselves in the foot by using these


duetosymmetry:
Just saw that the commit message on ee9d7c3 refers to derivatives instead of integrals. Oops!

duetosymmetry:
@moble @10220 any thoughts about including this?

moble:
Thanks.

PR #24, opened 2019-10-22
Bugfix: _WaveformBase__num needs to be included for pickling

duetosymmetry:
Pickling and unpickling a WaveformBase object doesn't work.  The
problem resides in the class variable `WaveformBase.__num`, which gets
name mangled to `_WaveformBase__num` because of python conventions.  The
current `__getstate__`/`__setstate__` pair try to avoid copying
`_WaveformBase__num` in order to preserve the auto-incremented counter
`__num`, which is automatically set in `__init__`. However, pickle does
something funny that I don't fully understand, so `_WaveformBase__num`
ends up not being a member unless it *is* copied. Minimal breaking
example below.

The fix is to let `__init__` auto-increment `__num`; then upon entering
`__setstate__`, store this value in a temp, update the `__dict__`, which
overwrites `__num`, but then revert `__num` to the temp.

Minimal breaking example which is fixed by this bugfix:

```python
import scri
import pickle
W = scri.WaveformModes()
W_str = pickle.dumps(W)
W2 = pickle.loads(W_str)
print('_WaveformBase__num' in W.__dict__.keys())
print('_WaveformBase__num' in W2.__dict__.keys())
```

### DIFF
--- a/waveform_base.py
+++ b/waveform_base.py
@@ -754,8 +754,11 @@ class WaveformBase(_object):
         back to a quaternion array.
 
         """
-        old_num = state.pop('_WaveformBase__num')
+        new_num = self.__num
+        old_num = state.get('_WaveformBase__num')
         self.__dict__.update(state)
+        # Make sure to preserve auto-incremented num
+        self.__num = new_num
         self.frame = quaternion.as_quat_array(self.frame)
         self._append_history('copied, deepcopied, or unpickled as {0}'.format(self), 1)
         self._append_history('{0} = {1}'.format(self, '{0}'.format(self).replace(str(self.num), str(old_num))))


duetosymmetry:
@moble are you comfortable merging this and issuing a new release on conda/pypi?

moble:
I've actually been looking into this a little, and I'm worried about it.  If `self` is correctly initialized on entry to this function, your new code should actually work the same as the existing code, because of the old use of `pop` instead of your `get`.  So I'm worried that this is just sticking a band-aid over one little problem and ignoring a more fundamental one that could have worse effects.

Also, your email from June says that 

> `state.pop('_WaveformBase__num')` fails during `waveform.copy()`. This throws a `KeyError`...

But `pop` should fail if and only if `get` will also fail.  One solution to *that* problem would just be to add a trivial default, like `state.pop('_WaveformBase__num', -1)`.  But your `get` would also need a default, or you'd see the same `KeyError`.  

I don't understand what's happening here, and I think we need to before sweeping the problem under the rug.

duetosymmetry:
Ok sorry to not address the previous issue. That previous issue arose if you pickled, and then tried to `.copy()` (or use other pieces of the code that explicitly look for `.__dict__['_WaveformBase__num']`. Those are *side effects* of this more fundamental problem, which is that `__setstate__` seems to create an object that's missing an entry in `__dict__`.

duetosymmetry:
Rebased onto upstream

moble:
Wow, this is weird behavior.  So on entry to `__setstate__`, the object has a `__num` attribute, which can be read and set, even though the `__dict__` attribute is empty, and everywhere else, `__num` and `__dict__['_WaveformBase__num']` are indistinguishable.  Bizarre.  I should have just used a single underscore.

Anyway, this is the only variable that needs to be treated this way, so I'm happy to merge.

Thanks!

duetosymmetry:
I also wish I understood better... maybe an underlying bug in the way that objects with inheritance are deepcopied? But I don't have the depth of understanding needed to track that down.

moble:
I can see why this would behavior a good thing, to save time in unpickling, but it just adds to the aura of black magic surrounding the leading double underscore... but not the leading-and-trailing double underscore...

PR #25, opened 2019-11-23
Fix mislabeled elements in repr

10220:
When printing a WaveformModes object, the data was misaligned with the labels. This correctly aligns the data with the correct label.


moble:
Thanks!

Issue #26, opened 2019-12-06
Tests fail due to pytest deprecating direct fixture call


10220:
Directly calling a fixture has been deprecated in pytest. Thus any tests that call `*_waveform` fixtures will fail. It seems like the fix is either to just remove all arguments from `*_waveform` or get _really_ clever with the pytest parametrizations.

10220:
This is addressed by PR #27

PR #27, opened 2019-12-11
Make fixture-generating funcs work again after pytest deprectation

duetosymmetry:
See
https://docs.pytest.org/en/latest/deprecations.html#calling-fixtures-directly
for details on the deprecation. The fixture-generating functions:
  constant_wavform
  linear_waveform
  random_waveform
  delta_waveform
were designed to be general and take optional arguments. Only their
default version should be considered as fixtures, because otherwise
they need to be called directly.  But, pytest has deprecated direct
calling of fixtures.  Therefore we make 0-argument fixtures (with
explicit names given to the decorator) which return the default values
from the above fixture-generating functions.  Then the underlying
functions can still be called with optional arguments.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -23,7 +23,6 @@ def pytest_runtest_setup(item):
         pytest.skip("Need `--run_slow_tests` command-line argument to run")
 
 
-@pytest.fixture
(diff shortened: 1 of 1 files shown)


10220:
This addresses issue #26

moble:
Thanks, Leo!

PR #28, opened 2019-12-11
Fix sqrt in extrapolation

10220:
Extrapolation was still failing even with commit ae82f6e08bd951955ebbec4cfe4b66dfe9b20db0. The error is:
```
TypeError: ufunc 'sqrt' not supported for the input types, and the inputs could not be safely coerced to any supported types according to the casting rule ''safe''
```

### DIFF
--- a/mode_calculations.py
+++ b/mode_calculations.py
@@ -551,7 +551,7 @@ def corotating_frame(
             *np.mean(Vhat_corot, axis=0)
         ).normalized()
         # print(i1, i2, i1m, i1p, RoughDirection, Vhat[0], Vhat_corot[0], Vhat_corot_mean)
-        correction_rotor = np.sqrt(-quaternion.z * Vhat_corot_mean).inverse()
+        correction_rotor = (-quaternion.z * Vhat_corot_mean).sqrt().inverse()
     # R = squad(R, t, W.t)
     if return_omega:
         return (frame * correction_rotor, omega)


moble:
I'm a little worried about this.  What is `numpy.version.version` in the numpy you're getting the error from?  Can you try the following code with that version as well?

```python
import numpy as np
import quaternion
np.sqrt(quaternion.x)
np.sqrt(np.array([quaternion.y, quaternion.z]))
```

That code works fine for me on 1.17.1 and 1.17.4 (which is current), but it looks like one or both of those last lines would fail for you.

10220:
Ah, both lines fail for me in both numpy 1.17.3 and 1.17.4

Let me try reinstalling quaternion

10220:
Okay so for some reason `conda update quaterion` failed to get me the correct version of quaternion but when I did:
```
conda uninstall quaterion
conda install -c conda-forge quaternion
```
Then this issue is fixed and everything works. I think I may have been pulling from `-c moble` by mistake. When I try to install quaternion from there then this fails. I'll close the PR.

10220:
For posterity, is there a reason why np.sqrt(quaterion.x) is prefered over quaternion.x.sqrt()?

moble:
> I think I may have been pulling from `-c moble` by mistake.

I've pretty much stopped using my own anaconda channel in favor of conda-forge, so you should probably remove my channel from your `~/.condarc` if it's still in there.  FYI, I have
```
channel_priority: strict
channels:
  - conda-forge
  - defaults
  - numba
```
because this works best for my purposes.  Of course, I also use miniconda and a conda env other than `base` as my default.  YMMV.

> For posterity, is there a reason why np.sqrt(quaterion.x) is prefered over quaternion.x.sqrt()?

Mostly consistency.  There are a lot of weird edge cases that result from numpy not fully supporting custom objects, but they're generally avoided if you can just use `np.sqrt`.  Either one should work in this particular case, but if the code gets changed at all (or used as a template for subtly different code), it could easily introduce awful bugs that depend very precisely on input types, array sizes and shapes, etc., because of python's flexibility with types.

Anyway, your error mostly made me worry that I didn't understand the new approach to ufuncs, which could hide other bugs.  Given that it works with an up-to-date quaternion, I have a mild preference for sticking with the current code, so I'll close this now.

10220:
Got it, thanks for clarifying that for me!

Sorry for wasting your time with this :see_no_evil:

PR #29, opened 2019-12-12
Extrapolation test

10220:
Two main things going on in this PR:

* Several of the tests require sample files in order to be run. I've added a script that will download two small waveforms from the catalog. Users can still use their own files to run the tests if they want.
* A test for waveform extrapolation for both default scri output and NRAR output.


10220:
Hey @moble, I just went back through this branch to confirm that nothing is outdated. This PR is ready to be looked at. Since the extrapolation test actually doesn't take too much time I don't restrict it to `--run_slow_test`.

10220:
This PR has had a big overhaul. We have removed the script for downloading a waveform from the catalog. Instead, we have added a new function that will generate a fake `rh_FiniteRadii_CodeUnits.h5` file (with a lot of flexibility for other uses) and the tests run with that. However, I have also set up `test_file_io` to use the output of `test_NRAR_extrapolation`. To ensure this dependency, both functions need to be in the same file and in the correct order.

10220:
Pardon those last two fixups, just a spelling mistake. It's all clean and ready to review now!

tests/test_SpEC.py
```diff
@@ -9,28 +9,90 @@
 import h5py
 import scri
 import scri.SpEC
+import shutil
```
moble: This can go.

tests/test_SpEC.py
```diff
@@ -9,28 +9,90 @@
...
+    """
+    output_dir = output_dir_base + "test_extrapolation_output/"
+    # Extrapolation fails if the output filenames exist so we delete
+    # the output files from a previous test if they exist.
+    if os.path.exists(output_dir):
+        shutil.rmtree(output_dir)
+
+    scri.extrapolation.extrapolate(
+        InputDirectory=input_dir, OutputDirectory=output_dir, ChMass=1.0
+    )
```
moble: Then, you can use the fixture like this:

```suggestion
def test_extrapolation(tempdir):
    """Test extrapolation of waveforms with output in the scri format"""
    input_dir = str(tempdir)
    output_dir = str(tempdir / "test_extrapolation")
    scri.extrapolation.extrapolate(InputDirectory=input_dir, OutputDirectory=output_dir, ChMass=1.0)
```
Note that there's no need to check if the input file exists; it will because the fixture created it.  No need to remove the output dir if it exists; it won't because the fixture created a new temporary directory.

Basically the same setup can be used for the next function, with `"test_extrapolation"` -> `"test_NRAR_extrapolation"`.

.gitignore
```diff
@@ -57,7 +57,9 @@ target/
 # PyCharm
 .idea
 
+test_output
 rhOverM_Asymptotic*h5
+rh_FiniteRadii_CodeUnits.h5
```
moble: These changes can be reverted if you use my suggestion of temporary directories.

tests/test_SpEC.py
```diff
@@ -9,28 +9,90 @@
...
+        ChMass=1.0,
+    )
+
+
+test_file = "../SpEC/samples/test_output/test_NRAR_extrapolation_output/rhOverM_Extrapolated_N2.h5"
+
+@pytest.mark.skipif(
+    not os.path.exists(test_file), reason="requires {}".format(test_file)
+)
 def test_file_io():
```
moble:
```suggestion
def test_file_io(tempdir):
```

tests/test_SpEC.py
```diff
@@ -9,28 +9,90 @@
...
     """Test file I/O with a round-trip and compare H5 files"""
-    w = scri.SpEC.read_from_h5('../SpEC/samples/rhOverM_Asymptotic_GeometricUnits.h5/Extrapolated_N2.dir')
-    scri.SpEC.write_to_h5(w, '../SpEC/samples/Asymptotic_GeometricUnits_test.h5/Extrapolated_N2.dir')
+    w = scri.SpEC.read_from_h5(test_file)
+    scri.SpEC.write_to_h5(w, "../SpEC/samples/Asymptotic_GeometricUnits_test.h5")
```
moble: This idea that the previous test must have run before this one is dangerous.  It's safer to just create the extrapolated file if it doesn't exist.

```suggestion
    input_file = tempdir / "test_NRAR_extrapolation" / "rhOverM_Extrapolated_N2.h5"
    output_dir = tempdir / "test_file_io"
    output_file = str(output_dir / "Asymptotic_GeometricUnits.h5")
    output_file_result = str(output_dir / "rhOverM_Asymptotic_GeometricUnits.h5")

    if not input_file.exists():
        scri.extrapolation.extrapolate(
            InputDirectory=str(tempdir),
            OutputDirectory=str(tempdir / "test_NRAR_extrapolation"),
            DifferenceFiles="",
            PlotFormat="",
            UseStupidNRARFormat=True,
            ChMass=1.0,
        )
    output_dir.mkdir()

    w = scri.SpEC.read_from_h5(input_file)
    scri.SpEC.write_to_h5(w, output_file)
```

tests/test_SpEC.py
```diff
@@ -9,28 +9,90 @@
...
 
-    f1 = h5py.File('../SpEC/samples/rhOverM_Asymptotic_GeometricUnits.h5')
-    f2 = h5py.File('../SpEC/samples/rhOverM_Asymptotic_GeometricUnits_test.h5')
-    w1 = f1['Extrapolated_N2.dir']
-    w2 = f2['Extrapolated_N2.dir']
+    f1 = h5py.File(test_file,'r')
+    f2 = h5py.File("../SpEC/samples/rhOverM_Asymptotic_GeometricUnits_test.h5",'r')
```
moble: These should have been context managers:

```python
    with h5py.File(input_file, "r") as f1, h5py.File(output_file_result, "r") as f2:
```

tests/test_SpEC.py
```diff
@@ -9,28 +9,90 @@
...
 
 
-@pytest.mark.skipif(not os.path.exists('../SpEC/samples/rhOverM_Asymptotic_GeometricUnits.h5'),
-                    reason="requires SpEC/samples/*.h5")
+# NOTE: test_file_io() depends on output from test_NRAR_extrapolation(), so the latter
+# must come first in this file.
+
+filename = "rh_FiniteRadii_CodeUnits.h5"
+input_dir = "../SpEC/samples/"
+output_dir_base = "../SpEC/samples/test_output/"
```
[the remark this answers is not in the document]
10220: 1. This commit suggestions feature might be one of the handiest things I've seen.
2. Ah, I was trying to fool around a bit with the fixtures but I think setting the scope to "session" was a big piece I was missing.

tests/test_SpEC.py
10220: If we are going to not require the previous test to be run first, then let's move the extrapolation tests to it's own file `test_extrapolation.py` so that it matches the organization of the rest of the tests. I'll do this in a followup commit right after.

moble: There's no need to *require* the extrapolation to run again; if the extrapolation test has already run, it's a nice time-saver to skip redoing it.  Plus, if you move it, you'll need to repeat or factor out the fixture.  And the extrapolation code is specific to SpEC format, so I don't see any real need to separate them.

If you want, you could pull out this call to `scri.extrapolation.extrapolate` as a mini function, since it's just repeated in the two tests with the same arguments.

10220: True, okay I now shift my platform to proposing that we move `extrapolation.py` into the `scri/SpEC/` subdirectory. Let's leave `test_SpEC.py` as it is, and since the extrapolation tests do come first in the file they will run first (I'd still like to leave a comment so that in the future we don't reorder them needlessly). But now I'm thinking that `extrapolation.py` would be better organized specifically in `SpEC`.

moble: Yeah, that sounds reasonable.  Just remember to change `__init__.py` and `SpEC/__init__.py` to import it from its new location.

moble:
Actually, do you know where the current extrapolation script actually used by SpEC people resides?  And if so, can you check that moving it won't break imports?

10220:
Yes, I had to work with Mark back when we changed the script from `GWFrames` to `scri`. I'll make sure I test it to make sure everything works.

10220:
Actually, I'm going to shift my platform yet again. To avoid overly complicating things, it might just be better to wait until we do the reorganization of the entire code. This would be better for SXS and for other users of scri as well that might be using functions inside `extrapolation.py` besides `extrapolate()`.

tests/test_SpEC.py
```diff
@@ -11,26 +11,78 @@
...
+
+
+def test_extrapolation(tempdir):
+    """Test extrapolation of waveforms with output in the scri format"""
+    input_dir = str(tempdir)
+    output_dir = str(tempdir / "test_extrapolation")
+    scri.extrapolation.extrapolate(InputDirectory=input_dir, OutputDirectory=output_dir, ChMass=1.0)
+
+
+def test_extrapolation(tempdir):
```
moble:
```suggestion
def test_NRAR_extrapolation(tempdir):
```

sample_waveforms.py
```diff
@@ -506,3 +506,165 @@ def sigma(ell):
...
+            R = areal_radius[0,1]
+            tortoise_coord = R + 2 * initial_adm_energy * np.log(R / (2 * initial_adm_energy) - 1)
+            # Compute the approximate time in SpEC simulation coordinates
+            simulation_time = (h0.t + tortoise_coord) * np.sqrt(1 - 2 * initial_adm_energy / R) / avg_lapse
+            for l in range(2, ell_max + 1):
+                for m in range(-l, l + 1):
+                    index = sf.LM_index(l, m, 2)
+                    new_data = h0.data[:, index]
+                    for n in range(1, n_subleading + 1):
+                        new_data += amp * h0.abs[:, index] * R ** -n * np.exp((1j * n * 50 * np.pi / h0.n_times) * h0.t)
```
moble: I'm sure the difference is teensy here, so there's no need to change it, but FYI: the reason I put `h0.t` and `h0.abs[:, index]` at the ends of their respective expressions is because python performs operations left-to-right.  So if you put all the floats at the beginning, you're doing as many (fast) float operations as possible, and minimizing the number of (slow) array operations.  I've gotten significant speedups from that in some cases.  Just something to keep in mind when dealing with larger arrays in core functions.

moble:
Okay, other than `test_NRAR_extrapolation` everything looks good to me.  Once that's done, are you ready to merge?

sample_waveforms.py
10220: :scream: How did I not know this?! I'll throw this in with the change you've requested below, it takes two seconds.

10220:
Do you want me to squash or are you going to do your own clean up of the commits on your end?

moble:
I have a really fancy trick that will let me do the squashing:

image

moble:
Thanks for this!

PR #30, opened 2020-02-28
Add BMS transformation for all Weyl scalars

10220:
Adds extends the functionality of the BMS transformation to all the Weyl scalars. Allows the user to select which eth convention to use.

### DIFF
--- a/waveform_grid.py
+++ b/waveform_grid.py
@@ -3,7 +3,8 @@
 
 from __future__ import print_function, division, absolute_import
 
-from . import (Inertial, WaveformModes, SpinWeights, h, sigma, psi0, psi1, psi2, psi3)
+from . import Inertial, WaveformModes, SpinWeights, DataNames
+from . import h, hdot, sigma, news, psi0, psi1, psi2, psi3, psi4
 from .waveform_base import WaveformBase, waveform_alterations
 
 import sys
@@ -303,7 +304,9 @@ class WaveformGrid(WaveformBase):
           3. Boost
         All input parameters refer to the transformation required to take the mode's inertial frame onto the inertial
         frame of the grid's inertial observers.  In what follows, the inertial frame of the modes will be unprimed,
-        while the inertial frame of the grid will be primed.
+        while the inertial frame of the grid will be primed.  NOTE: These are passive transformations, e.g. supplying 
+        the option space_translation=[0, 0, 5] to a Schwarzschild spacetime will move the coordinates to z'=z+5 and so 
+        the center of mass will be shifted in the negative z direction by 5 in the new coordinates.
 
         The translations (space, time, spacetime, or super) can be given in various ways, which may override each
         other.  Ultimately, however, they are essentially combined into a single function `alpha`, representing the
@@ -350,6 +353,21 @@ class WaveformGrid(WaveformBase):
         boost_velocity : float array of length 3, optional
             This is the three-velocity vector of the grid frame relative to the mode frame.  The norm of this vector
             is checked to ensure that it is smaller than 1.  Defaults to [], corresponding to no boost.
+        psi4_modes : WaveformModes, required only if w_modes is type psi3, psi2, psi1, or psi0
+        psi3_modes : WaveformModes, required only if w_modes is type psi2, psi1, or psi0
+        psi2_modes : WaveformModes, required only if w_modes is type psi1 or psi0
+        psi1_modes : WaveformModes, required only if w_modes is type psi0
+            The objects storing the modes of the original waveforms of the same spacetime that 
+            w_modes belongs to. A BMS transformation of an asymptotic Weyl scalar requires mode 
+            information from all higher index Weyl scalars. E.g. if w_modes is of type scri.psi2, 
+            then psi4_modes and psi3_modes will be required. Note: the WaveformModes objects 
+            supplied to these arguments will themselves NOT be transformed. Please use the 
+            AsymptoticBondiData class to efficiently transform all asymptotic data at once.
+        eth_convention : {'SXS', 'NP', 'GHP'}
+            Specify the convention for the spin-weight raising/lower operator. Options are:
+                * 'SXS', the SXS collaboration convention (equal to eth_NP/2)
+                * 'NP', the Newman-Penrose convention
+                * 'GHP', the Geroch-Held-Penrose convention (equal to eth_NP/np.sqrt(2))
 
         Returns
         -------
@@ -484,10 +502,6 @@ class WaveformGrid(WaveformBase):
         gamma = 1 / math.sqrt(1 - beta**2)
         varphi = math.atanh(beta)
 
-        if kwargs:
-            import pprint
-            warnings.warn("\nUnused kwargs passed to this function:\n{0}".format(pprint.pformat(kwargs, width=1)))
-
         # These are the angles in the transformed system at which we need to know the function values
         thetaprm_j_phiprm_k = np.array([[[thetaprm_j, phiprm_k]
                                          for phiprm_k in np.linspace(0.0, 2*np.pi, num=n_phi, endpoint=False)]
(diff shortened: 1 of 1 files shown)


10220:
Would you like me to add a test to this, in this PR or as a separate PR?

moble:
I think we can wait on the tests.  It'll be easier once we have this and AsymptoticBondiData in the same branch.

moble:
I'm struggling with this 'eth_convention' parameter.

Setting aside the moral imperative against introducing new conventions without very strong reasons, and the questionable name "SXS" (I don't know that anyone in SXS has actually used this convention before), I think the definition of ð should be regarded as an implementation detail that must remain invisible to the user.  In particular, nothing the user inputs is affected by the choice of a convention for ð.  Basically, the Weyl components rely on choices made for the metric signature, the definition of Riemann, the tetrad components, and the signs of the components themselves — none of which know anything about ð.<sup>(*)</sup>  The BMS group is, of course, independent of ð.  And since it transforms to another coordinate system where the Weyl components again know nothing about ð, the output of `from_modes` should also be unaffected by the choice of a convention for ð.  So I think the right approach is just to choose one (preferably pre-existing) ð convention and stick with it.

On the other hand, there certainly are different choices for the tetrad — which does affect the input and output.  The only important differences I'm seeing are in the definition of `m`, which can differ by factors of -1 and/or √2.  As I mentioned before, I think that -1 explains the sign differences between my version of the transformation equations and those of Gómez López and Quiroga, for example.  Their choice also differs from mine by a factor of √2, which seems to be absorbed *in their equations* by their different choice of normalization for ð.  (They also made some other mistakes, but let's ignore them.)  So comparing our equations is really comparing apples to oranges; if you were to just pick one convention for ð and write the equations with that choice, the differences wouldn't be down to just the signs, but also factors of √2 — *both* of which should be explained by the difference in `m` alone.

So I claim that you should be able to ignore different conventions for ð; just pick one, write down your equations, and incorporate different conventions for `m` into this function.

---

<sup>(*)</sup> Decomposing the waveforms into SWSH mode weights relies on the SWSHs themselves, but there are conventions for those independent of what one chooses for ð.  Anyway, I've never seen any mention of conventions for SWSHs that differ by anything other than `(-1)**m`, and there are very explicit conventions for SWSHs in NR, so I think we can ignore this.

moble:
I forgot to mention n and l.  To be explicit, here are the various relations — as far as I can tell — between the asymptotic tetrads as used by SXS, Moreschi/me, and Gómez López-Quiroga:



But since they all use the same (at least up to an overall sign) expressions for the Weyl components in terms of the tetrad, this means that there are non-uniform relationships between the Weyl components as calculated in the different systems, and thus between the transformation laws as written in the different systems.  If my algebra is correct, we have — up to that overall sign — the following relations:



and maybe even this, if there are no further surprises:



If you copy the transformation equations as derived in my paper or GLQ's paper, for use with SXS conventions, you'd need to substitute accordingly (on both sides of the equations).  Similarly with the Bianchi identities, etc.

I think it's fair to just use SXS conventions, and assume that the input waveforms use those conventions.  If it ever comes up that people are using different conventions, the easiest thing to do would just be to convert the input and output waveforms themselves rather than the transformation equations.  On the other hand, if you want to directly support different conventions, it would probably be best to take as input parameters the various tetrad factors relative to the SXS conventions.  (Though for now, I think it's reasonable to just support SXS conventions until we're more confident that these transformations actually work with them.)

10220:
I definitely agree with your first comment but I'm having a bit of a hard time reconciling the dictionary you have in your second comment with my own. I'll work through it again and see if that illuminates anything.

moble:
I did leave out the factors of exp(-iϕ) from m_Moreschi and m_GLQ — just because those should cancel out in the transformation equations — and I'll certainly admit the possibility of overall signs missing from the Weyl components.  Are those enough to explain the differences?

moble:
Oops.  I meant factors of exp(iϕ).

10220:
Unfortunately, it's not just signs. It's still a factor of sqrt(2) :cry:

10220:
I'm just looking at how the SpEC waveforms satisfy the Bianchi identites and I'm varying the factors in front of each term in the equations. There are two things that seem to be a source of error:
1. Since {Ψ2, Ψ1, Ψ0} are extracted on an inner set of strahkorpers, they have a different set of corrected u than {h, Ψ4, Ψ3}
2. The strain h is extracted via a completely different (and more sensitive to noise) method. Thus the terms multiplying conj(h) and a Weyl scalar could be particularly problematic.

Here we finally have a stroke of luck! A clean way to test how h performs is by looking at the relative errors of the two equations:
Ψ3 = ðḣ
Ψ˙3 = ðΨ4

The following plot shows the relative error and we can see that there is a massive difference.
RWZ_vs_WCE

10220:
After extracting the NP shear, σ, I found that this dictionary (which doesn't quite make sense) is able to convert the SpEC waveforms to be used in Moreschi Equations:
Screenshot from 2020-03-11 17-14-30

There are still a few issues that I'm trying to work out now.

10220:
Actually a better variant of that dictionary that makes more sense is one in which the minus signs on the tetrad vectors (ell,n) are instead on (m,mbar).

moble:
Wait, I thought you said the other day that in SpEC Ψ₄ ≈ -2σ̈ .  This could make a lot more sense if that were true.

10220:
That was true on a first glance, but after I improved the extraction, ran with higher resolution, and got a clean extrapolation the factor is ~2.828 = sqrt(8). This still agrees with the dictionary because the two m^a vectors and the ell^a vector will all throw a factor of sqrt(2) in the equation for σ, which all ends up as sqrt(8). However, I'm going back through the procedure to make sure I didn't forget anything.

10220:
We are computing σ in the simulation via:
Screenshot from 2020-03-11 18-14-14

where t_a = (-N,0,0,0) is the timelike unit normal to the spatial hypersurfaces, N is the lapse, Γ^a_{bc} are the spacetime Christoffel symbols, and R_a is the unit normal to the strahlkorper.

10220:
I ended up finding the bug that was leading to a strange ψ₃ waveform, and I also found another bug where the shift terms weren't contributing to the electric Weyl tensor term in the Weyl tensor construction (This shouldn't make much a difference at all at ~100 M, but this might explain some of the deviation between the explicit extraction and characeristic field extraction). I'm rerunning this now with Lev1 only so I will have results to discuss on Friday. I've attached the PDF that details what we have done so far.

[wheres_the_factor.pdf](https://github.com/moble/scri/files/4383547/wheres_the_factor.pdf)

10220:
Alright I have enough of a waveform to say that, as expected, adding in the missing shift terms only made the Weyl tensor waveforms and the Weyl characteristic fields waveforms agree better. After fixing the bug in ψ₃ the waveforms also agree for that as well.

10220:
Following a discussion between Mike, Saul, and me, I went about testing the Bianchi identity between ψ3 and ψ4 for Weyl scalars extracted with explicitly defined tetrads and without decomposing them into modes. So out of the simulation, we get the value for these Weyl scalars on a grid of points at a fixed coordinate radius. 

I take the time derivative of ψ3 by applying a spline derivative, and then I take the data from a single time `t_i`. At the time `t_i`, I take the eth derivative of ψ4 by taking spline derivatives along θ and φ directions via the follwing code:
```
n_θ = n_φ = 2*ell_max+1
ðψ4 = np.zeros((n_θ, n_φ), dtype=complex)
for θ_i in range(n_θ):
    for φ_i in range(n_φ):
        dψ4_dθ = InterpolatedUnivariateSpline(θ_range, ψ4[t_i,   :, φ_i].real).derivative()(θ_range) + 1j*InterpolatedUnivariateSpline(θ_range, ψ4[t_i,   :, φ_i].imag).derivative()(θ_range)
        dψ4_dφ = InterpolatedUnivariateSpline(φ_range, ψ4[t_i, θ_i,   :].real).derivative()(φ_range) + 1j*InterpolatedUnivariateSpline(φ_range, ψ4[t_i, θ_i,   :].imag).derivative()(φ_range)
        ðψ4[θ_i, φ_i] = -(dψ4_dθ[θ_i] + 1j*dψ4_dφ[φ_i]/np.sin(θ_range[θ_i]) + 2*ψ4[t_i,θ_i,φ_i]/np.tan(θ_range[θ_i])) / np.sqrt(2)
```
Aaaaaaaaand \*drumroll\* we still find the factor of sqrt(2)...
bianchi_on_grid

10220:
Now the Pen & Paper & Mathematica approach:

We use the metric and Riemann tensor for the M=0 even parity Teukolsky wave, c.f. [Teukolsky 1982](https://journals.aps.org/prd/abstract/10.1103/PhysRevD.26.745). The metric is (-,+++) and we are working in the TT gauge.

We define the strain, tetrad, and Weyl scalars as given in Appendix C of the [2019 SXS Catalog Paper](https://arxiv.org/abs/1904.04831).

With this, we find that indeed to satisfy the (linear) asymptotic Bondi gauge Bianchi identities appear to be:
Screenshot from 2020-03-30 16-53-37

These are satisfied to machine precision in Mathematica at various values of θ (values off-axes as well).

10220:
After some analytic work, here is a full update of this convention issue so far. It presents the work done up to this point and compares our conventions explicity against the Moreschi-Boyle convention.
[wheres_the_factor.pdf](https://github.com/moble/scri/files/4423361/wheres_the_factor.pdf)

10220:
@moble this PR is now ready to go. For the sake of posterity I will explicitly mention here that this BMS transformation will only work with SpEC convention waveforms. We will address this convention issue after this PR and the Bondi Data Container PR are merged.

PR #31, opened 2020-06-11
Exclude insignificant outer radii from Psi1 and Psi0 extrapolation

10220:
SpEC's extraction procedure has been optimized for Psi1 and Psi0 waveforms. This involves having the innermost extraction radius moved to a MUCH smaller radius (2 reduced GW wavelengths as determined by the initial data). The outermost extraction has not changed. 

Since Psi1 and Psi0 fall off as r^-4 and r^-5, the amplitude of even their dominant mode will fall beneath the error tolerance of SpEC within the extraction region. To optimize the extrapolation for these quantities, we find which outer radii on a given time slice will have a waveform value beneath a user-set tolerance. These outer radii will then be excluded from being used in extrapolation.

### DIFF
--- a/extrapolation.py
+++ b/extrapolation.py
@@ -1433,7 +1433,7 @@ def _Extrapolate(FiniteRadiusWaveforms, Radii, ExtrapolationOrders, Omegas=None,
     NExtrapolations = len(ExtrapolationOrders)
     SVDTol = 1.0e-12  # Same as Numerical Recipes default in fitsvd.h
     DataType = FiniteRadiusWaveforms[NFiniteRadii-1].dataType
-    ExcludeInsignificantRadii = True if ((DataType in [scri.psi1, scri.psi0]) and ErrorTol) else False
+    ExcludeInsignificantRadii = DataType in [scri.psi1, scri.psi0]) and bool(ErrorTol)
     if ExcludeInsignificantRadii:
         from spherical_functions import LM_index
         ell_min = FiniteRadiusWaveforms[NFiniteRadii - 1].ell_min
--- a/extrapolation.py
+++ b/extrapolation.py
@@ -1433,7 +1433,7 @@ def _Extrapolate(FiniteRadiusWaveforms, Radii, ExtrapolationOrders, Omegas=None,
     NExtrapolations = len(ExtrapolationOrders)
     SVDTol = 1.0e-12  # Same as Numerical Recipes default in fitsvd.h
     DataType = FiniteRadiusWaveforms[NFiniteRadii-1].dataType
-    ExcludeInsignificantRadii = DataType in [scri.psi1, scri.psi0]) and bool(ErrorTol)
+    ExcludeInsignificantRadii = DataType in [scri.psi1, scri.psi0] and bool(ErrorTol)
     if ExcludeInsignificantRadii:
         from spherical_functions import LM_index
         ell_min = FiniteRadiusWaveforms[NFiniteRadii - 1].ell_min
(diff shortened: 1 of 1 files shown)


extrapolation.py
```diff
@@ -1421,6 +1432,12 @@ def _Extrapolate(FiniteRadiusWaveforms, Radii, ExtrapolationOrders, Omegas=None)
     NFiniteRadii = len(FiniteRadiusWaveforms)
     NExtrapolations = len(ExtrapolationOrders)
     SVDTol = 1.0e-12  # Same as Numerical Recipes default in fitsvd.h
+    DataType = FiniteRadiusWaveforms[NFiniteRadii-1].dataType
+    ExcludeInsignificantRadii = True if ((DataType in [scri.psi1, scri.psi0]) and ErrorTol) else False
```
moble, line 1436:
```suggestion
    ExcludeInsignificantRadii = DataType in [scri.psi1, scri.psi0]) and bool(ErrorTol)
```
Note that `ErrorTol is not None` is the usual test, but `bool(ErrorTol)` also handles cases where `ErrorTol==0`.

10220: Thanks for the idea about the bool cast, I hadn't thought of that.

10220:
Hey @moble, I know you have a 1e6 things on your plate but I just want to make sure this PR doesn't fall through the cracks. Some of the runs are starting to finish on Caltech-HPC and it would be fun to start extrapolating them and seeing what we have.

extrapolation.py
```diff
@@ -591,6 +591,14 @@ def extrapolate(**kwargs):
...
           of the waveform: (W_outer.T(0)+W_outer.T(-1))/2
 
+        ErrorTol                 None
+          ONLY USED FOR Psi1 AND Psi0 WAVEFORMS.
+          ErroTol is the effective noise floor of the simulation.
+          If Psi1 or Psi0 (NOT scaled by radius) falls beneath
+          this tolerance for certain extraction radii, then those
+          radii are not included in the extrapolation. The value
+          1.0e-9 seems to work well for a few BBH systems.
+
```
moble, line 601: I didn't understand what this meant until I read the code.  Calling it a tolerance makes me think that the code will run some procedure until it achieves some level of accuracy, and makes me think that it might be a fractional rather than an absolute value.  I think renaming this variable `NoiseFloor` and removing the word "tolerance" would make it clearer.

extrapolation.py
```diff
@@ -1561,6 +1578,29 @@ def _Extrapolate(FiniteRadiusWaveforms, Radii, ExtrapolationOrders, Omegas=None)
...
+                # Since we are in the corotating frame, we can be sure that the (2,2) mode is dominant. For the sake of
+                # consistency we will use the same radial weights for each mode at a given retarded time.
+                WaveformAmplitude = np.abs(Re[MinimumNRadii,LM_index(2,2,ell_min)] + 1j*Im[MinimumNRadii,LM_index(2,2,ell_min)])
+
+                # Radius at which the amplitude of the waveform is equal to error_tol.
+                LargestSignificantRadius = (WaveformAmplitude/ErrorTol)**(1/(6-DataType))
+
+                # Use as many radii as possible that are smaller than LargestSignificantRadius.
+                RadiiOnTimeSlice = np.array(Radii)[:,i_t]
+                MinimumNRadii= max(len(RadiiOnTimeSlice[RadiiOnTimeSlice <= LargestSignificantRadius]), MinimumNRadii)
```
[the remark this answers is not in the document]
10220: I had no idea you could sum an array of bools without any casts! Python is wild.

10220:
Hey @moble, thanks for the review and for catching some oversights! I committed your suggestions and added another commit to address renaming `ErrorTol` to `NoiseFloor`. 

I just tested this latest version of the code and everything works.

moble:
Thanks!

PR #32, opened 2020-06-13
Update README.md

10220:
I just noticed that we never updated the conda install from `moble` to `conda-forge`.


moble:
Good catch.  Thanks.

PR #33, opened 2020-06-14
Docs building on readthedocs

duetosymmetry:
Currently enabled on my fork, will transfer once this PR is merged.

### DIFF
--- a/.readthedocs.yml
+++ b/.readthedocs.yml
@@ -1,12 +1,12 @@
 version: 2
 
+conda:
+  environment: docs/environment.yml
+
 python:
-  version: 3.7
-  system_packages: true
   install:
     - method: pip
       path: .
-    - requirements: requirements-doc.txt
 
 # Don't build any extra formats
 formats: []
--- a/.readthedocs.yml
+++ b/.readthedocs.yml
@@ -1,0 +1,12 @@
+version: 2
+
+python:
+  version: 3.7
+  system_packages: true
+  install:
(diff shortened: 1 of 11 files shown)


moble:
This is awesome.  Thanks so much for doing this!

duetosymmetry:
The webhook now has to be enabled on your upstream repo (moble/scri) instead of my repo. I don't think there is a way to 'transfer' it over manually. Instead it's probably easiest if we delete the RTD project, and you 'import' the project from your RTD>github integration.

I gave you admin access to my RTD scri project, so you can delete it when you're ready (I think! Let me know if it doesn't let you delete it). I think everything should work with default config settings when you import, since the `.readthedocs.yml` file overrides all the relevant stuff. The only stuff that will be lost are the values I entered for:

- Description
  Python/numba code for manipulating time-dependent functions of spin-weighted spherical harmonics
- Tags
  general-relativity, numerical-methods, physics, python, scientific-computing

but of course feel free to change these. I don't even know where this metadata shows up.

BTW I think some of the python docstrings are not in official `numpydoc` format, which is why they don't look quite right in the auto-built docs.

moble:
Okay, it looks like it's working now.  I just removed the old webhook and added a new one.  Thanks again!

duetosymmetry:
Ok, I have deleted the github side of things on my repo. Hopefully we'll see that the next commit to your repo successfully triggers a rebuild on RTD.

PR #37, opened 2020-07-31
Set ABD internal conventions to Moreschi-Boyle

10220:
This just sets the conventions for the constraint equations to Moreschi-Boyle for the `AsymptoticBondiData` class. A later PR will introduce a function to load waveforms into an `ABD` object and convert between conventions.

### DIFF
--- a/scri/asymptotic_bondi_data/transformations.py
+++ b/scri/asymptotic_bondi_data/transformations.py
@@ -310,8 +310,9 @@ def transform(self, **kwargs):
     # variation in u, the second axis variation in θ', and the third axis variation in ϕ'.
     u = self.u
     α = sf.Grid(supertranslation.evaluate(distorted_grid_rotors), spin_weight=0).real[np.newaxis, :, :]
-    ðα = sf.Grid(supertranslation.eth.evaluate(distorted_grid_rotors), spin_weight=α.s + 1)[np.newaxis, :, :]
-    ððα = sf.Grid(supertranslation.eth.eth.evaluate(distorted_grid_rotors), spin_weight=α.s + 2)[np.newaxis, :, :]
+    # The factors of 1/sqrt(2) and 1/2 come from using the GHP eth instead of the NP eth.
+    ðα = sf.Grid(supertranslation.eth.evaluate(distorted_grid_rotors)/np.sqrt(2), spin_weight=α.s + 1)[np.newaxis, :, :]
+    ððα = sf.Grid(0.5*supertranslation.eth.eth.evaluate(distorted_grid_rotors), spin_weight=α.s + 2)[np.newaxis, :, :]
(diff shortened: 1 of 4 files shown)


10220:
Hey @moble, thanks for pointing out the failing tests, I had forgotten to change the signs in `from_initial_values.py`. In regards to our conversation earlier, I agree that there could be a good motivating decision to pick a different sign convention, but I would like to advocate for the Moreschi-Boyle convention because there is very nice documentation through the literature alraedy but also it nicely supports all of the papers that `scri` was intended for originally. Just my two cents for whatever it's worth :man_shrugging:

scri/asymptotic_bondi_data/transformations.py
```diff
@@ -310,8 +310,9 @@ def transform(self, **kwargs):
     # variation in u, the second axis variation in θ', and the third axis variation in ϕ'.
     u = self.u
     α = sf.Grid(supertranslation.evaluate(distorted_grid_rotors), spin_weight=0).real[np.newaxis, :, :]
-    ðα = sf.Grid(supertranslation.eth.evaluate(distorted_grid_rotors), spin_weight=α.s + 1)[np.newaxis, :, :]
-    ððα = sf.Grid(supertranslation.eth.eth.evaluate(distorted_grid_rotors), spin_weight=α.s + 2)[np.newaxis, :, :]
+    # The factors of 1/sqrt(2) and 1/2 come from using the GHP eth instead of the NP eth.
+    ðα = sf.Grid(supertranslation.eth.evaluate(distorted_grid_rotors)/np.sqrt(2), spin_weight=α.s + 1)[np.newaxis, :, :]
+    ððα = sf.Grid(0.5*supertranslation.eth.eth.evaluate(distorted_grid_rotors), spin_weight=α.s + 2)[np.newaxis, :, :]
```
moble, line 315: When I read things like this, where a different function is used in one or two lines, it always slows me down as I try to figure out if there's some difference.  Is there some reason not to do this?

```suggestion
    ðα = sf.Grid(supertranslation.eth_GHP.evaluate(distorted_grid_rotors), spin_weight=α.s + 1)[np.newaxis, :, :]
    ððα = sf.Grid(supertranslation.eth_GHP.eth_GHP.evaluate(distorted_grid_rotors), spin_weight=α.s + 2)[np.newaxis, :, :]
```

10220: Unforutnately, `eth_GHP` is a `ModesTimeSeries` object but `supertranslation` is a `spherical_functions.Modes` object which doesn't support `eth_GHP`. I chose to just include the factors here to minimize amount of code but another option is to promote `supertranslation` to a `ModesTimeSeries` object. I can do that if you prefer?

moble: Oh, I see.  No, this is better.

PR #38, opened 2020-08-01
Load h5 files into an ABD object

10220:
[**DEPENDS ON PR #37** - The convention dictionary here assumes `constraints.py` is in Moreschi-Boyle conventions] 

This provides a convenient way to load H5 files into an ABD object. It can handle various file types and can be easily extendible for other waveform conventions as well. This is just one way to handle the issue of differing sign conventions so I'm open for other thoughts or discussion!

### DIFF
--- a/scri/SpEC/__init__.py
+++ b/scri/SpEC/__init__.py
@@ -5,4 +5,4 @@
 from .._version import __version__
 
 from .com_motion import com_motion, estimate_avg_com_motion, remove_avg_com_motion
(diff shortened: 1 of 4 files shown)


10220:
Hey @moble, I've added a significant addition to this PR. I've added the feature of having an ABD with dimensionless data since that is what we work with. Now the BMS transformation for ABD works with our waveforms.

10220:
I'm going to close this PR, these commits will instead be included in a future PR adding more functionality to ABD.

PR #42, opened 2020-08-20
Fix failing tests

10220:
Hey @moble, here is a fix for the failing tests. A few `eth`s didn't get updated to `eth_GHP`.


moble:
Thanks.  I don't understand how that passed the tests before, either...  Anyway, they pass now.

PR #43, opened 2020-09-23
Replace an eth_NP with eth_GHP

10220:
We forgot one eth_NP -> eth_GHP replacement.


10220:
Hey @moble, I've updated the test so that everything passes.

PR #51, opened 2021-01-21
Use Github Actions to automate testing with pytest

10220:
This will automatically run our tests with pytest whenever a commit is pushed. It will also run on Pull Requests. I'm not sure if the azure pipeline is required for how scri is deployed to conda-forge so I left that alone for now.

In this PR I've also fixed a few bugs that `flake8` caught and I trim up some trailing whitespace.


moble:
This looks good, though I'm not sure why it didn't run the action.  Maybe it's being blocked by the older CI stuff?  Anyway, this would be a good time to remove that: the directories

  - .azure-pipelines
  - .ci_support
  - .circleci
  - .scripts
  - recipe

and the files

  - azure-pipelines.yml
  - conda-forge.yml

10220:
@moble I added a commit that removes the old testing infrastructure but the new tests still are not running. I can confirm that the new tests work on [my fork](https://github.com/10220/scri/tree/new_testing_pipeline). I guess it needs to be merged before the tests will actually run? :man_shrugging:

moble:
Worth a shot...

moble:
Yep.  It's started now.  Thanks!

PR #52, opened 2021-01-27
Fix error message about insufficient number of finite-radius objects

hrueter:
When you run extrapolation with barely sufficient number of radii the following error warning was triggered.
`ERROR: Asking for extrapolation up to order 4, but only got 5finite-radius Waveform objects; need at least 5 waveforms.`

This PR fixes the consistency check.


moble:
I'm not saying I approve of using the minimal number, but you're right that this shouldn't have raised an error.  Thanks!

PR #59, opened 2021-09-29
Adding BMS frame fixing code

keefemitman:
This PR predominantly adds the file map_to_bms_frame.py in the asymptotic_bondi_data directory, which includes two functions, `transformation_to_map_to_CoM_frame` and `transformation_to_map_to_BMS_frame` which can be used to fix the frame of an ABD object with either a target strain waveform or the Moreschi supermomentum.

(I know this could use some work, but I just wanted to put something out there so I could hear your (@moble) thoughts on how to improve it and what else I should include so we can add this to scri)

### DIFF
--- a/scri/asymptotic_bondi_data/__init__.py
+++ b/scri/asymptotic_bondi_data/__init__.py
@@ -208,3 +208,7 @@ class AsymptoticBondiData:
         bondi_CoM_charge,
         supermomentum,
     )
+    from .map_to_bms_frame import (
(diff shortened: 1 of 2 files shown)


moble:
Thanks.  I've started taking a look, and I'll try to finish an initial review next week.

keefemitman:
Please hold off on reviewing this! I'm working on a new code that should find all supertranslations up to some `ell_max` and should do so much more quickly (using an iterative scheme).

PR #60, opened 2021-10-01
Extrapolation can read rpxmb finite-radius files.

markscheel:
It still can read nrar files as before.
I rewrote most of the code for reading files, using new
functions that already exist like scri.SpEC.file_io.read_from_h5

### DIFF
--- a/scri/extrapolation.py
+++ b/scri/extrapolation.py
@@ -203,112 +203,144 @@ def validate_group_of_waveforms(h5file, filename, WaveformNames, LModes):
         print("In '{}', the following waveforms are not valid:\n\t{}".format(filename, "\n\t".join(FailedWaveforms)))
     return Valid
 
+def read_finite_radius_waveform_nrar(filename, WaveformName):
 
-def read_finite_radius_waveform(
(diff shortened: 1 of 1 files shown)


markscheel:
This now passes the following tests:
 - Extrapolating nrar files agrees with the old method of extrapolating nrar files.
 - Extrapolating rpxmb files agrees (to 1e-10) with extrapolating nrar files.

Notes:
 - There is still an argument LModes in read_finite_radius_data that is not really used anymore, except for validation of nrar files.  Should I get rid of LModes?
 - Validation of nrar files is still done exactly as before.  There is no validation of rpxmb files, because presumably those are created from nrar files that have already been validated.

moble:
> * There is still an argument LModes in read_finite_radius_data that is not really used anymore, except for validation of nrar files. Should I get rid of LModes?

I don't mind.  If it's getting in the way, feel free to remove it.

> * Validation of nrar files is still done exactly as before. There is no validation of rpxmb files, because presumably those are created from nrar files that have already been validated.

That sounds fine by me.  The validation was crucial back in the days when we routinely had missing segments, non-monotonic data, etc.  But I haven't heard any reports of such problems on this end of the pipeline in years.

PR #79, opened 2022-08-27
Adding Lorentz/BMS_transformation classes and tests

keefemitman:
This includes code for `Lorentz_transformation` and `BMS_transformation` classes. For each of these two classes, you can reorder, invert, and compose transformations. The motivation behind this is for when we might want to easily transition between the superrest frame and the PN BMS frame, e.g., for future surrogate modeling. The main ideas behind this code are from Spinors and Spacetime, Volume 1, Chapters 1.2/1.3 as well as the latter parts of DOI:10.1063/1.1705135.

(The main parts of this code work, but there are parts I plan on improving after I return from vacation on September 5th).

@moble @duetosymmetry


scri/bms_transformations.py
```diff
@@ -0,0 +1,829 @@
...
+            
+    return conformal
+
+def nCr(n, r):
+    """n choose r.
+    """
+    if n >= 0 and r >= 0 and n - r >= 0:
+        return np.math.factorial(n)/(np.math.factorial(r)*np.math.factorial(n-r))
+    else:
+        return 0
```
duetosymmetry, line 269: Expanding out each factorial and then taking the product and ratio can lead to overflow or just loss of precision. `scipy.special.binom` and `scipy.special.comb` use methods to keep the numerator and denominator close to each other and minimize the number of multiplications, to minimize loss of precision.

scri/bms_transformations.py
```diff
@@ -0,0 +1,829 @@
...
+    if ell_max is None:
+        ell_max = Lorentz.ell_max
+    n_theta = 2 * ell_max + 1; n_phi = n_theta;
+        
+    transformed_S = transform_function(S, Lorentz, ell_max, tol)
+    
+    transformed_S_S2 = spinsfast.salm2map(transformed_S, 0, ell_max, n_theta, n_phi)
+    K_S2 = compute_conformal_factor(Lorentz, n_theta, n_phi)
+
+    S_prime_S2 = K_S2**(-1)*transformed_S_S2
```
duetosymmetry, line 408: Why isn't this just:
```suggestion
    S_prime_S2 = transformed_S_S2 / K_S2
```

scri/bms_transformations.py
```diff
@@ -0,0 +1,829 @@
...
+        if not ('rotation' in output_order and\
+                'boost' in output_order):
+            raise ValueError('Not enough transformations')
+        
+        if self.order == output_order:
+            return self
+        else:
+            L_spin_matrix = Lorentz_to_spin_matrix(self)
+            L_reordered = spin_matrix_to_Lorentz(L_spin_matrix, output_order=output_order)
+            return L_reordered
```
duetosymmetry, line 468: If the order matches, we get back the original object, whereas if the order differs, we get back a new object. If we get back the same object, mutating the result will mutate the original; but not if we get back a different object. Maybe the first case should make a copy of itself instead of return self?

scri/bms_transformations.py
```diff
@@ -0,0 +1,829 @@
...
+
+    def is_identity(self, verbose=False):
+        """Check if a BMS transformation is the identity element
+        """
+        rotation_is_identity = np.allclose(self.rotation.components, [1,0,0,0])
+        boost_is_identity = np.allclose(self.boost, [0,0,0])
+        if verbose:
+            print(f"rotation is identity: {rotation_is_identity}")
+            print(f"boost is identity: {boost_is_identity}")
+        return rotation_is_identity and boost_is_identity
```
duetosymmetry, line 478: Other functions are taking a tolerance argument, but this one is not?

scri/bms_transformations.py
```diff
@@ -0,0 +1,829 @@
...
+        output_order: list
+            Order in which rotation and boost should be applied.
+        """
+        L1_spin_matrix = Lorentz_to_spin_matrix(self)
+        L2_spin_matrix = Lorentz_to_spin_matrix(other)
+        
+        conjugate_spin_matrix = np.linalg.multi_dot([L2_spin_matrix, L1_spin_matrix])\
+            - np.linalg.multi_dot([L1_spin_matrix, L2_spin_matrix])
+    
+        return spin_matrix_to_Lorentz(conjugate_spin_matrix, output_order=output_order)
```
duetosymmetry, line 529: This looks like it's computing the commutator of spin matrices --- is that really the same as one Lorentz transformation conjugated by a second? I thought that rather it would have been L2 . L1 . inv(L2) . The documentation should also say what exactly is meant by the conjugation of one Lorentz transformation by another.

scri/bms_transformations.py
```diff
@@ -0,0 +1,829 @@
...
+    """
+    # Maybe better to define Pauli matrices?
+    a = np.cos(fourvec[0] / 2) + 1j * fourvec[3] * np.sin(fourvec[0] / 2)
+    b = (-fourvec[2] + 1j * fourvec[1]) * np.sin(fourvec[0] / 2)
+    c = (+fourvec[2] + 1j * fourvec[1]) * np.sin(fourvec[0] / 2)
+    d = np.cos(fourvec[0] / 2) - 1j * fourvec[3] * np.sin(fourvec[0] / 2)
+
+    return np.array([[a, b], [c, d]])
+
+def Lorentz_to_spin_matrix(Lorentz):
```
duetosymmetry, line 119: I thought the pythonic convention was that instances of classes should still start with a lower case letter (though the name of the class itself starts with an upper case letter)

scri/bms_transformations.py
```diff
@@ -0,0 +1,829 @@
...
+    """Convert a Lorentz transformation to a spin matrix.
+    
+    Parameters
+    ----------
+    Lorentz: Lorentz_transformation
```
duetosymmetry, line 124: It looks like the code below assumes that the argument represents a Lorentz transformation with a certain `order` -- doesn't check the `order` field to determine if that's what's meant. Maybe the order field should be tested, and reorder if not in the presumed order.

scri/bms_transformations.py
```diff
@@ -0,0 +1,829 @@
...
+    else:
+        boost_vec_hat = Lorentz.boost \
+            / np.linalg.norm(Lorentz.boost)
+        
+    # Compute product of Lorentz vector and Pauli matrices
+    rotation_spin_matrix = fourvec_to_spin_matrix([psi, *rotation_vec_hat])
+    boost_spin_matrix = fourvec_to_spin_matrix([chi, *boost_vec_hat])
+
+    if Lorentz.order.index('rotation') < Lorentz.order.index('boost'):
+        return np.linalg.multi_dot([boost_spin_matrix, rotation_spin_matrix])
```
duetosymmetry, line 147: When it's only 2 matrices, we don't really need `multi_dot`

scri/bms_transformations.py
```diff
@@ -0,0 +1,829 @@
...
+    thetaprm_phiprm = sf.theta_phi(n_theta, n_phi)
+
+    [[a, b], [c, d]] = Lorentz_to_spin_matrix(Lorentz)
+
+    conformal = np.zeros((n_theta, n_phi), dtype=complex)
+    for j in range(n_theta):
+        for k in range(n_phi):
+            thetaprm_j, phiprm_k = thetaprm_phiprm[j, k]
+            if thetaprm_j == 0:
+                z = np.inf
```
duetosymmetry, line 255: This value of z is not used

scri/bms_transformations.py
```diff
@@ -0,0 +1,829 @@
...
+        for p in range(min(Lmax-M1, L, L + M) + 1):
+            sum_contribution += (-1)**(p-M)*nCr(L,p)*nCr(L,p-M)*nCr(Lmax-L,Lmax-M1-p)
+        return ALM/np.sqrt(float(np.math.factorial(L)*np.math.factorial(L)))*sum_contribution
+    
+def transform_Y_LM(Y_LM, Lorentz, ell_max=None):
+    """Apply a Lorentz transformation to a spherical harmonic.
+
+    Parameters
+    ----------
+    f_LM: ndarray, dtype=complex
```
duetosymmetry, line 308: Docs parameter name doesn't match what's above. I think this is actually an array-like with two (real) elements, meaning (L,M) -- is that right?

scri/bms_transformations.py
```diff
@@ -0,0 +1,829 @@
...
+        ell_max = Lorentz.ell_max
+
+    f_LM_prime = np.zeros((ell_max + 1)**2, dtype=complex)
+    for L in range(0, int(np.sqrt(f_LM.size) - 1) + 1):
+        for M in range(-L, L+1):
+            if abs(f_LM[sf.LM_index(L, M, 0)]) < tol:
+                continue
+            f_LM_prime += f_LM[sf.LM_index(L, M, 0)] * transform_Y_LM((L, M), Lorentz, ell_max)
+
+    return f_LM_prime
```
duetosymmetry, line 377: I feel like this function is doing way more work than it needs to... every Y_{\ell m} is being separately transformed on a grid and the transformed data converted back from position space to harmonic space... and then summing over all \ell m. But I feel that it must be possible to do this with only one transformation, rather than (\ell+1)^2 transformations.

scri/bms_transformations.py
```diff
@@ -0,0 +1,829 @@
...
+        L_inverse = spin_matrix_to_Lorentz(L_spin_matrix_inverse, output_order=output_order)
+        return L_inverse
+
+    def __eq__(self, other):
+        rotation_eq = np.allclose(self.rotation.components, other.rotation.components)
+        boost_eq = np.allclose(self.boost, other.boost)
+        return rotation_eq and boost_eq
+            
+    def compose(self, other, output_order=['rotation','boost']):
+        """Compose two Lorentz transformations.
```
duetosymmetry, line 497: I think it would be useful here to say whether this is left- or right-multiplication, i.e. is it self*other or other*self, when each is viewed as a (passive) Lorentz transformation on the coordinates.

scri/bms_transformations.py
```diff
@@ -0,0 +1,829 @@
...
+
+def _process_transformation_kwargs(input_ell_max, **kwargs):
+    original_kwargs = kwargs.copy()
+
+    # Build the supertranslation and spacetime_translation arrays
+    supertranslation = np.zeros((4,), dtype=complex)  # For now; may be resized below
+    ell_max_supertranslation = 1  # For now; may be increased below
+    if "supertranslation" in kwargs:
+        supertranslation = np.array(kwargs.pop("supertranslation"), dtype=complex)
+        if supertranslation.dtype != "complex" and supertranslation.size > 0:
```
duetosymmetry, line 16: Did you want:
```suggestion
        if not (supertranslation.dtype == "complex" and supertranslation.size > 0):
```

scri/bms_transformations.py
```diff
@@ -0,0 +1,829 @@
...
+            # I don't actually think this can ever happen...
+            raise TypeError(
+                "Input argument `supertranslation` should be a complex array with size>0.  "
+                f"Got a {supertranslation.dtype} array of shape {supertranslation.shape}"
+            )
+        # Make sure the array has size at least 4, by padding with zeros
+        if supertranslation.size <= 4:
+            supertranslation = np.lib.pad(
```
duetosymmetry, line 24:
```suggestion
            supertranslation = np.pad(
```

scri/bms_transformations.py
```diff
@@ -0,0 +1,829 @@
...
+            for m in range(ell + 1):
+                i_pos = sf.LM_index(ell, m, 0)
+                i_neg = sf.LM_index(ell, -m, 0)
+                a = supertranslation[i_pos]
+                b = supertranslation[i_neg]
+                if abs(a - (-1.0) ** m * b.conjugate()) > 3e-16 + 1e-15 * abs(b):
+                    raise ValueError(
+                        f"\nsupertranslation[{i_pos}]={a}  # (ell,m)=({ell},{m})\n"
+                        + "supertranslation[{}]={}  # (ell,m)=({},{})\n".format(i_neg, b, ell, -m)
+                        + "Will result in an imaginary supertranslation."
```
duetosymmetry, line 47:
```suggestion
                        + "Will result in a complex supertranslation."
```

scri/bms_transformations.py
```diff
@@ -229,45 +229,36 @@ def spin_matrix_to_Lorentz(A, output_order=['rotation','boost']):
         boost_idx = np.inf
         
     if rotation_idx < boost_idx:
-        rotation_matrix = np.linalg.multi_dot([u, vh])
+        rotation_matrix = np.matmul([u, vh])
```
duetosymmetry, line 232:
```suggestion
        rotation_matrix = np.matmul(u, vh)
```

scri/bms_transformations.py
```diff
@@ -0,0 +1,725 @@
...
+    Parameters
+    ----------
+    A: 2 by 2 complex array
+        2 by 2 array corresponding to the input spin matrix.
+    output_order: list
+        Order in which rotation and boost should be applied.
+    """
+    if np.allclose(A, np.zeros_like(A)):
+        return Lorentz_transformation()
+
```
duetosymmetry, line 219: Should this be if A is the identity or minus the identity matrix?

duetosymmetry:
TODO: Also update the code in WaveformGrid. The code for computing the transformed grid is common to both WaveformGrid and AsymptoticBondiData. Should this be factored out?

tests/test_bms_transformations.py
```diff
@@ -0,0 +1,450 @@
...
+
+    Lorentz_composed = Lorentz2.compose(Lorentz1)
+
+    abd1 = abd.transform(frame_rotation=Lorentz1.rotation.components, boost_velocity=Lorentz1.boost)
+
+    abd2 = abd1.transform(frame_rotation=Lorentz2.rotation.components, boost_velocity=Lorentz2.boost)
+
+    abd_composed = abd.transform(
+        frame_rotation=Lorentz_composed.rotation.components, boost_velocity=Lorentz_composed.boost
+    )
```
moble, line 248: See, this is tricky.  It says that applying `L1` to `abd`, then applying `L2` to the transformed `abd` is the same as applying `L1 * L2`.  Put another way: if we think of `L1` and `L2` as *operators* on the waveform, we would write that as `L2(L1(abd)) = (L1*L2)(abd)`.  So composition is opposite to multiplication: `L1*L2 = L2∘L1`

The reason, of course, is that `L2` is expressed in the first frame, so to apply it to `L1(abd)`, you have to transform it into the second frame — which is `L1*L2*L1.inverse()` — so `L2∘L1 = L1*L2*L1.inverse()*L1 = L1*L2`.

This is the subtlety I was worried about in our phone call yesterday.  And I agree that you've handled it correctly here, but just reading the documentation of `compose`, I think it would be easy to order things the wrong way around.

scri/asymptotic_bondi_data/transformations.py
```diff
@@ -145,9 +145,8 @@ def Bprm_j_k(thetaprm, phiprm):
     for j in range(n_theta):
         for k in range(n_phi):
             thetaprm_j, phiprm_k = thetaprm_phiprm[j, k]
-            R_j_k[j, k] = (
-                Bprm_j_k(thetaprm_j, phiprm_k) * frame_rotation * quaternion.from_spherical_coords(thetaprm_j, phiprm_k)
-            )
+            rotated_coord_quaternion = frame_rotation * quaternion.from_spherical_coords(thetaprm_j, phiprm_k)
+            R_j_k[j, k] = Bprm_j_k(*quaternion.as_spherical_coords(rotated_coord_quaternion)) * rotated_coord_quaternion
```
moble, line 149: I agree that this is the right thing to do.  Thanks for finding this!

moble (changes requested):
I think it's easier to see that the `frame_rotation` in `abd.transformations` was wrong with a simpler test that doesn't involve any of the new machinery.  Specifically, `test_asymptoticbondidata.py` should have a test that just applies a rotation `R`, then applies a boost `v`, and compares that to applying the transformation with `R` and `R.inverse()*v*R` simultaneously.  Probably another one that does the same, but with a supertranslation first.

scri/asymptotic_bondi_data/transformations.py
keefemitman: No problem! Glad we agree.

scri/bms_transformations.py
```diff
@@ -0,0 +1,799 @@
+import copy
+import scipy
+import spinsfast
+from scri.asymptotic_bondi_data.transformations import *
+import numpy as np
+import spherical_functions as sf
+
+
+def _process_transformation_kwargs(input_ell_max, **kwargs):
```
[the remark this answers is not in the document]
keefemitman: Yeah, will do!

scri/bms_transformations.py
```diff
@@ -0,0 +1,799 @@
...
+    if lorentz.order.index("rotation") < lorentz.order.index("boost"):
+        return np.matmul(boost_spin_matrix, rotation_spin_matrix)
+    else:
+        return np.matmul(rotation_spin_matrix, boost_spin_matrix)
+
+
+def Lorentz_to_quaternion(lorentz):
+    a, b, c, d = Lorentz_to_spin_matrix(lorentz).flatten()
+    print((a + d) / 2, (b + c) / (2j), (-b + c) / 2, (a - d) / (2j))
+    return np.quaternion((a + d) / 2, (b + c) / (2j), (-b + c) / 2, (a - d) / (2j))
```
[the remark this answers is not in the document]
keefemitman: I mean to delete this code as I never ended up using it anyways (sorry about that).

scri/bms_transformations.py
```diff
@@ -0,0 +1,799 @@
...
+        )
+
+        if bms_inverse.order == output_order:
+            return bms_inverse
+        else:
+            return bms_inverse.reorder(output_order=output_order)
+
+    # Functions involving two BMS transformations
+
+    def __eq__(self, other):
```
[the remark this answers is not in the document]
keefemitman: This was mainly for when I was testing things. I'll remove it now that things seem to be working.

tests/test_bms_transformations.py
keefemitman: I'll try and clean up the documentation. I also wrote up some TeX that Leo edited that we think would be useful to add to the documentation of `scri` to make things clearer (https://www.overleaf.com/6964495134vvzdnkwnwkhv)

duetosymmetry: We should continue adding details here — namely, we didn't include the math of conjugating group elements by other elements for the composition of two BMS elements.

duetosymmetry: I just updated the Overleaf with that abstract calculation, but not the details of how conjugating works.

keefemitman:
What's the best way to incorporate this .TeX file into the docs? https://www.overleaf.com/6964495134vvzdnkwnwkhv

keefemitman:
@moble do you want me to write the test function for the boosts that we talked about?

moble:
Thanks @keefemitman and @duetosymmetry

PR #80, opened 2022-08-31
Center of mass correction for matter

markscheel:
Various changes to CoM correction to handle matter:
 * Now handles NSNS as well as BHNS
 * Reads correct fields from Matter.h5 (which was standardized after the previous CoM version)
 * Does not assume Horizons.h5 and Matter.h5 have the same timesteps, but **does** assume that the first N timesteps (for some N) of Matter.h5 and Horizons.h5 are the same, and it cuts off subsequent timesteps.  This is because Matter.h5 is not always output at late times, e.g. for BHNS where the BH swallows the NS.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -47,9 +47,8 @@ jobs:
         if: ${{ env.skipping_build_and_test_replicate != 'true' }}
         shell: bash
         run: |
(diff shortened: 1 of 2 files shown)


moble:
Can you "Allow edits from maintainers" please?

Issue #81, opened 2022-09-21
Should ABD objects optimize for when some fields are omitted?


duetosymmetry:
We may be in the situation where we use and abd object to store only strain, or just strain and \psi_i through \psi_4 with some smallest value of i. In this case, there are a lot of calculations in e.g. `AsymptoticBondiData.transform()` that could be skipped. For example, in the lines https://github.com/moble/scri/blob/6db96adae014591a03bcbce9b80bc3d26377ff20/scri/asymptotic_bondi_data/transformations.py#L325-L336
we can simply skip all the fields which are not stored; and similarly in the lines
https://github.com/moble/scri/blob/6db96adae014591a03bcbce9b80bc3d26377ff20/scri/asymptotic_bondi_data/transformations.py#L342-L387
we can skip the calculations for fields \psi_j where j<i, i representing the smallest subscript of Weyl scalar we keep, as above.

If I remember correctly, @akhadse has implemented this as a speedup. Was that right, Akshay?

@moble do you think this is a feature worth having for ABD objects?

akhadse:
@duetosymmetry - Yes that is right. I check there if self.fields_present[i]==True and skip the calculations for the corresponding field if they are not present

moble:
I think the objective is good, but it would pay to be thoughtful about the implementation.  In particular, it's dangerous to start putting in flags and branches, because it increases code complexity, and we would need to make sure that those flags are supported in every piece of code.  Plus, it makes it harder to ensure that what we write reflects the mathematics.  I think there's a much easier solution that should be more generic and elegant.

[...]

PR #82, opened 2022-11-29
Reduce memory usage when applying supertranslations by a factor of 6

dongzesun:
Applying the supertranslations to the abd quantities one by one to reduce memory usage.


dongzesun:
Thanks for the review. But I also need to find the translations that map to the PN BMS frame, so at least psi_2 is needed apart from the strain. Should I write another function that only transforms psi_2 and strain?

duetosymmetry:
Tangentially related to #81

PR #85, opened 2023-02-10
Update code to use BMSTransformation class and add function to map ABD object to the frame of another

keefemitman:
This PR updates the `map_to_superrest_frame.py` code to use the class `scri.bms_transformations.BMSTransformation` to keep track of the BMS transformations that we're applying. By doing so, this enables us to only ever transform the original ABD object (since we can compose transformations), which, in theory, should help reduce numerical noise.

Apart from updating the `map_to_superrest_frame.py` code, this also adds a new file, `map_to_abd_frame.py`, which includes the function `map_to_abd_frame`. This function maps one ABD object to the frame of another.

This is achieved by doing the following:
* find the transformation that maps `target_ABD` to the superrest frame; call it `T1`
* find the transformation that maps `self` to the superrest frame; call it `T2`
* find the transformation that time/phase aligns the strains in the superrest frame; call it T_time_phase
* construct the map from `self` to `target_ABD` via `T1.inverse() * T_time_phase * T2`
* perform another time/phase optimization by aligning the strains
* iterate this as many times as desired

To ensure this function works as expected, the test file `test_abd_frame.py` has also been added.

One thing I'm unsure of is how to define the error between the two ABD objects. If the user chooses to do an optimization over the time/phase freedom, then the error is simply the relative error between the strain waveforms. If not, then the error is defined to be the average of the relative error of the Weyl scalars and strain. I don't think this should matter too much, but if people have ideas that are better than what I've implemented, I'd be happy to change the code accordingly.

@moble @duetosymmetry might find this of interest.


keefemitman:
Oh I've also noticed that the tolerances in [line](https://github.com/moble/scri/blob/f49dcf5da3d5794060d0f07be8044091fd95c96d/scri/asymptotic_bondi_data/transformations.py#L43) are a bit stringent. @moble how did you come up with these? can we relax them?

scri/waveform_base.py
```diff
@@ -961,7 +961,7 @@ def interpolate(self, tprime):
         W.t = np.copy(tprime)
         W.frame = quaternion.squad(self.frame, self.t, W.t)
         W.data = np.empty((W.n_times,) + self.data.shape[1:], dtype=self.data.dtype)
-        W.data_2d[:] = CubicSpline(self.t, self.data_2d.view(float))(W.t).view(complex)
+        W.data_2d[:] = CubicSpline(self.t, self.data_2d.view(complex))(W.t).view(complex)
```
moble, line 964: I *think* I wrote this at a time when CubicSpline couldn't handle complex inputs.  But if it can nowadays, I suppose you can just remove both of these `.view`s, right?

scri/asymptotic_bondi_data/transformations.py
```diff
@@ -40,7 +40,7 @@ def _process_transformation_kwargs(input_ell_max, **kwargs):
                 i_neg = sf.LM_index(ell, -m, 0)
                 a = supertranslation[i_pos]
                 b = supertranslation[i_neg]
-                if abs(a - (-1.0) ** m * b.conjugate()) > 3e-16 + 1e-15 * abs(b):
+                if abs(a - (-1.0) ** m * b.conjugate()) > 1e-12 + 1e-12 * abs(b):
```
moble, line 43: IIRC, I chose these tolerances based on my experience, where I would explicitly impose symmetries.  E.g., I might just define the non-negative modes, and then fill in negative modes based on the reality requirement.

I worry about unexpected wierdness if the supertranslation is partially imaginary, even at the 1e-12 level.  I imagine somebody trying to diagnose a problem somewhere down the line, overlooking the cause because they quite reasonably assume supertranslations are always real.

Would it be better to arbitrarily just choose a way to impose reality?  Maybe we could take the appropriately signed "average" of the two modes and just reset each value?  If 1e-12 errors are not significant, what problems could it cause to impose reality like this?

scri/asymptotic_bondi_data/map_to_abd_frame.py
```diff
@@ -0,0 +1,478 @@
...
+    diff = integrate(
+        np.sum(
+            abs(modes_A(t_reference + δt_δϕ[0]) * np.exp(1j * δt_δϕ[1]) ** δϕ_factor * δΨ_factor - modes_B) ** 2, axis=1
+        ),
+        t_reference,
+    )[-1]
+    return np.sqrt(diff / normalization)
+
+
+def align2d(h_A, h_B, t1, t2, n_brute_force_δt=None, n_brute_force_δϕ=None, include_modes=None, nprocs=4):
```
moble, line 33: Ah, I remember now.  Another neglected PR to sxs has similar code.  If I merge that, could we just reuse it here?

duetosymmetry: I just realized that scri and sxs depend on each other... is this ok?

moble: `sxs` doesn't *really* depend on `scri`; it's optional, and I try to never use `scri` in `sxs`.  The reason is just that `scri` can be so hard to install, and `sxs` is intended to be this simple, pure-python package.  I think the only use is in [`version_info`](https://github.com/sxs-collaboration/sxs/blob/8f46c1c1c9c6de8ac69b69564d6a538934610547/sxs/utilities/__init__.py#L28), where it just sees if `scri` can be loaded, so it can record its version number; if not, it just skips it.  This sort of thing is fine in python, as long as you're not creating circular imports by putting `import scri` and `import sxs` at the top levels.  Basically, as long as one is hidden in a function, it's no problem.

On the other hand, I'm perfectly happy to use `sxs` from `scri`, and there are already instances of that.

scri/asymptotic_bondi_data/transformations.py
keefemitman: Yeah I think like the idea of a signed average of the two modes. I can try that and see if the tolerances are obeyed.

scri/asymptotic_bondi_data/map_to_abd_frame.py
keefemitman: Yeah I'd be fine with calling `sxs` to do the 2d alignment. Just lmk when that code is merged into `scri`!

keefemitman:
Looks like there are some sxs/scri compatibility issues that are causing some bugs. Working on a fix and will push sometime next week.

scri/asymptotic_bondi_data/bms_charges.py
```diff
@@ -99,6 +101,37 @@ def bondi_angular_momentum(self, output_dimensionless=False):
...
 
+def CWWY_angular_momentum(self):
+    """Compute the Chen/Wang/Wang/Yau angular momentum vector.
+
+    See Eq. (5) of https://arxiv.org/abs/2102.03235
+
+    """
+    ell_max = 1  # Compute only the parts we need, ell<=1
+    supertranslation_potential_data = scri.asymptotic_bondi_data.map_to_superrest_frame.𝔇inverse(
+        np.array(self.sigma.ethbar_GHP.ethbar_GHP + self.sigma.bar.eth_GHP.eth_GHP), self.ell_max
```
duetosymmetry, line 112: Is it faster to just compute `self.sigma.ethbar_GHP.ethbar_GHP` and take twice the real part?

PR #86, opened 2023-03-01
extrapolation.py: read rpdmb inputs.

markscheel:
Previously would read only rpxmb and nrar, now also can read rpdmb.

### DIFF
--- a/scri/extrapolation.py
+++ b/scri/extrapolation.py
@@ -278,14 +278,18 @@ def read_finite_radius_waveform_nrar(filename, WaveformName):
     
     return waveform,T,Indices,Radii,AverageLapse,CoordRadius,InitialAdmEnergy
 
-def read_finite_radius_waveform_rpxmb(filename, groupname, WaveformName):
+def read_finite_radius_waveform_rpxmb_or_rpdmb(filename,
+                                               groupname, WaveformName):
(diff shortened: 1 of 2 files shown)


markscheel:
I added a commit to stop reading Time.dat

scri/extrapolation.py
```diff
@@ -278,23 +278,56 @@ def read_finite_radius_waveform_nrar(filename, WaveformName):
...
+            read_rpxmb=True
+
+    # Note that groupname begins with a '/' so
+    # os.path.join(filename,groupname,WaveformName) does not work.
+    if read_rpxmb:
+        waveform=scri.rpxmb.load(filename+groupname+"/"+
+                                 WaveformName)[0].to_inertial_frame()
+    elif read_rpdmb:
+        waveform=scri.WaveformModes.from_sxs(
+            sxs.rpdmb.load(filename+groupname+"/"+WaveformName))
```
duetosymmetry, line 320: I think we need to update the dep ([in pyproject.toml](https://github.com/moble/scri/blob/f49dcf5da3d5794060d0f07be8044091fd95c96d/pyproject.toml#L19)) for `sxs` to be `>=2022.4.0` for `sxs.rpdmb.load`

markscheel: done

PR #88, opened 2023-08-24
Enabling python 11 and updated versions of tqdm

raffienficiaud:
This is a very basic PR

* bumping the allowed version of python to include python 11
* removing the upper version bound of tqdm to allow installation of eg. pandas 2

I am not very familiar with `poetry` but I am able to run this branch as part of a pip requirement file without issues.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -8,7 +8,7 @@ authors = ["Michael Boyle <michael.oliver.boyle@gmail.com>"]
 homepage = "https://github.com/moble/scri"
 
 [tool.poetry.dependencies]
-python = ">=3.8,<3.11"
+python = ">=3.8,<3.12"
 numpy = ">=1.20"
 scipy = "^1.0"
 numba = ">=0.55"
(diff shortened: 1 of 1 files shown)


moble:
Thanks.  Sorry for my slow reply, I had to remember which version of tqdm would actually work, because it was broken for a long time.

moble:
Yeah, I merged this (without deploying) so I could start a new branch to get rid of `poetry` in favor of the newer (and usually better) `hatch` build system, in hopes that it would succeed.  Of course, I've also upgraded to an M2 mac since the last time I built this, which means I have to try to get FFTW building, which is leading me down more rabbit holes...  Anyway, it's still a work in progress.

moble:
Those tests finally finished, after taking almost 4 hours to install with poetry.  My new branch installs (but not yet for macos ARM) in around 20 seconds.  So that's a bit better.

raffienficiaud:
```
Thu, 07 Sep 2023 15:22:41 GMT Resolving dependencies...
Thu, 07 Sep 2023 19:14:39 GMT
Thu, 07 Sep 2023 19:14:39 GMT Package operations: 83 installs, 0 updates, 0 removals
```

The problem seems to come from the number of possible combinations for the versions of the dependencies. The same will happen for `pip` for instance. I don't know how `hatch` solves this better.

For checking the compatibilities of versions, the package manager downloads the dependency to cache and checks the requirements (and sometimes it also needs to build that dependency if the `wheel` is not available). 
In the best case on the builders and when the `wheel` is available, the package manager downloads the candidate versions of the dependencies, which is usually the bottleneck. 
My guts feeling is that this is what is happening here, verbose logs should tell.

moble:
Sorry, I didn't mean to belabor the point with my last comment; I think this is resolved.  Poetry's slow dependency resolution is different from the one used by pip and hatch, and is a [known problem](https://github.com/python-poetry/poetry/issues/2094) that its devs are sticking their heads in the sand about, which is why I switched to hatch (which wasn't available last time I tried to drag my python distribution methods into the current decade).

PR #92, opened 2024-05-22
Update ``create_abd_from_h5`` to include other post-processing options; update documentation

keefemitman:
Update the ``create_abd_from_h5`` function to include other post-processing, i.e., time translate by the worldtube radius, scale out the total Christodoulou mass to make waveforms dimensionless, interpolate on to some coarser time array, and map to some reasonable BMS frame. Documentation in docs/tutorial_abd.rst has also been updated.

### DIFF
--- a/docs/tutorial_abd.rst
+++ b/docs/tutorial_abd.rst
@@ -74,7 +74,7 @@ relativists. First, the ABD class uses the Moreschi-Boyle convention, which is *
 the standard convention used in NR. See Appendices B and C of
 `arXiv:2010.15200 <https://arxiv.org/abs/2010.15200>`_ for more information. Second,
 the Newman-Penrose shear :math:`\sigma` is *not* the gravitational-wave strain :math:`h`.
-In the Moreschi-Boyle convention, we have the relation :math:`h = \bar{\sigma}`, but
+In the Moreschi-Boyle convention, we have the relation :math:`h = 2\bar{\sigma}`, but
 this relation is not gauranteed in every convention and only holds for asymptotic
 quantities.
 
@@ -282,16 +282,16 @@ composing BMS transformations. For more, see :ref:`bms_transformations`.
 BMS Frames
 ==========
 
-All waveforms at future null infinity (and all waveforms more generally) are functions or coordinates.
+All waveforms at future null infinity (and all waveforms more generally) are functions of coordinates.
 Therefore, there are certain "frames" which may be more useful than others, like that of a rest frame.
-For waveforms at future null infinity, the number of coordinates freedoms, i.e., the symmetries, that they
+For waveforms at future null infinity, the number of coordinate freedoms, i.e., the symmetries, that they
 exhibit is infinite and is summarized by a group known as the BMS group. This controls the types of frames
 that one may map waveforms to. Because GR is covariant, there is no preferred frame. However, for performing
 analysis on waveforms or building waveform models, it turns out that there are certain frames that are
 more useful than others. In particular, within GR one can extend the notion of a rest frame to something called
 a "superrest frame" (see arXiv:2405.08868 or arXiv:2208.04356 for more details), which typically yields waveforms
 that are easier to understand/analyze. Effectively, mapping to this frame amounts to mapping the system to be
-in the center-of-mass frame, with no instananeous memory, and it's angular velocity in the z-direction. For example,
+in the center-of-mass frame, with no instananeous memory, and its angular velocity in the z-direction. For example,
 for a remnant black hole, this corresponds to making the coordinates match those of the usual Kerr metric and is
 therefore incredibly useful (and necessary) for fitting QNMs to NR waveforms.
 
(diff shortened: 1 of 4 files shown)


nilsdeppe (commented):
This is really great! Thank you so much! I haven't yet tried to run it on CCE output, but here are some suggestions from a careful reading. Please just incorporate what you agree with and ignore what you don't :)

docs/tutorial_abd.rst
```diff
@@ -277,3 +277,76 @@ These components of a BMS transformation can also all be stored in the
...
+* ``target_strain_input``, the target strain; this should be ``None`` to map to the superrest frame,
+  but to map to the PN BMS frame one should input the PN strain (see arXiv:2208.04356).
+
+* ``padding_time``, the time window about ``t_0`` to be used when finding the BMS transformation to the superrest frame.
+
+========================================================
+Creating an AsymptoticBondiData Object from a CCE Output
+========================================================
+
+For processing the output of SpECTRE CCE, one may use the function ``scri.SpEC.file_io.create_abd_from_h5``.
```
knelli2, line 315: What's the difference between this and `scri.SpEC.create_abd_from_h5`?

docs/tutorial_abd.rst
```diff
@@ -277,3 +277,76 @@ These components of a BMS transformation can also all be stored in the
...
+  >>> import scri
+  >>> abd = scri.SpEC.file_io.create_abd_from_h5(
+        file_name='CharacteristicExtractVolume_R0292.h5',
+        file_format="spectrecce",
+        ch_mass=1.0,
+        t_interpolate=t_worldtube,
+        t_0_superrest=1600,
+        padding_time=200
+      )
+  >>> h = scri.asymptotic_bondi_data.map_to_superrest_frame.MT_to_WM(2.0*abd.sigma.bar)
```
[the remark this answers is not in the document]
knelli2: I think 2. of Nils' suggestions is probably the most important, because yeah I agree that people will just want to see a waveform when they are first trying this out. If they want to do more complicated things they'll have to read the papers and such.

I also agree there should be links to the docs on what a `WaveformModes` object is and does

knelli2 (changes requested):
I'll be linking to this exact page in the CCE tutorial

docs/tutorial_abd.rst
keefemitman: They're the same, `scri.SpEC` just imports `create_abd_from_h5` from `file_io`.

docs/tutorial_abd.rst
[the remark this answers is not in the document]
keefemitman: Ok I can just link to https://github.com/moble/scri/blob/main/docs/tutorial_waveformmodes.rst for 2.

scri/SpEC/file_io/__init__.py
```diff
@@ -581,4 +647,14 @@ def create_abd_from_h5(file_format, convention="SpEC", **kwargs):
         )
         abd.sigma = abd.sigma.bar
 
+    # Interpolate to finer time array, if specified
+    if not t_interpolate is None:
```
moble, line 651:
```suggestion
    if t_interpolate is not None:
```

scri/SpEC/file_io/__init__.py
```diff
@@ -581,4 +647,14 @@ def create_abd_from_h5(file_format, convention="SpEC", **kwargs):
...
+        idx1 = np.argmin(abs(t_interpolate - abd.t[0])) + 1
+        idx2 = np.argmin(abs(t_interpolate - abd.t[-1])) + 1 - 1
+        abd = abd.interpolate(t_interpolate[idx1:idx2])
+
+    # Map to superrest frame at some time over some window, if specified
+    if not (t_0_superrest is None or padding_time is None):
```
moble, line 657:
```suggestion
    if t_0_superrest is not None and padding_time is not None:
```

docs/tutorial_abd.rst
```diff
@@ -277,3 +277,76 @@ These components of a BMS transformation can also all be stored in the
...
+  E.g., to make reasonable-looking waveforms, one should map to the superrest frame at some time after junk;
+  ``t_0_superrest`` is the time at which to map to this frame, and ``padding_time`` is the window around ``t_0_superrest``
+  that is used when computing this BMS transformation. ``t_0_superrest - padding_time`` should be after junk radiation.
+  ``padding_time`` should be a few hundred :math:`h=2\overline{\sigma}`, e.g., two orbital periods.
+  The function used to do this is ``abd.map_to_superrest_frame`` (see the "BMS Frames" section).
+
+We recommend including all of these post-processing steps when processing a SpECTRE CCE output.
+
+To obtain the strain :math:`h` from the ``abd`` object, one can use the function ``scri.asymptotic_bondi_data.map_to_superrest_frame.MT_to_WM`` via ``h = MT_to_WM(2.0*abd.sigma.bar)``. 
+This is because the strain :math:`h` is related to the shear :math:`\sigma` via :math:`h=2\overline{\sigma}`.
```
moble, line 337: At some point we talked about adding a property `h` to ADB that would just do this for the user.  Do you want to do this now?

moble (changes requested):

docs/tutorial_abd.rst
moble: Do we want to raise it even higher to just `scri.create_abd_from_h5`, so as to not confuse people with the `SpEC` part?

knelli2: I agree, having `SpEC` in the name is a bit confusing when using SpECTRE CCE

scri/SpEC/file_io/__init__.py
```diff
@@ -429,18 +468,30 @@ def create_abd_from_h5(file_format, convention="SpEC", **kwargs):
 
     Parameters
     ----------
-    file_format : 'SXS', 'CCE', 'SpECTRECCE1', 'RPDMB', or 'RPXMB'
+    file_name : str,
+        Path to .h5 file to be loaded.
+    file_format : 'SXS', 'SpECTRECCE', 'RPDMB', or 'RPXMB'
```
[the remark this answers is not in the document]
nilsdeppe: Either's fine with me. I would say that the current format can reasonably be considered the first public version. Having said that, keeping the number the same between this and what we have inside the CCE output files could be helpful.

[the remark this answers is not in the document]
keefemitman: I was hoping we could just have the user specify `SpECTRECCE` and then the code could figure out how to extract data from the output file based on what version it is. The way I've set the code setup works for the previous version of SpECTRE CCE as well as the new version, and I figure we can just change it to work for whatever future versions there may be. Thoughts?

moble (changes requested):

scri/SpEC/file_io/__init__.py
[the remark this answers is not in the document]
moble: That's reasonable, but to ensure that doesn't fail in the future, we'd need to check that the input data actually has an `sxs_format = "SpECTRECCE_v1"` attribute, because we won't want this function to silently fail on `v2`.  (Then, to add `v2` support, we'll need to add new code anyway.)

knelli2 (commented):
I have no more comments. No need to wait for my review of the final changes to merge.

PR #93, opened 2024-06-02
Add 2nd monotonic requirement; fixes subtle bug affecting some CCE outputs

keefemitman:
@moble should be an easy review; fixes a bug that occurs when the output CCE data has time steps with differences ~1e-15, which are set to zero when doing `-= float(radius)` and screw up the monotonicity of the time array.

### DIFF
--- a/scri/SpEC/file_io/__init__.py
+++ b/scri/SpEC/file_io/__init__.py
@@ -604,10 +604,14 @@ def create_abd_from_h5(
         if file_format == "spectrecce_v1":
             WMs[i].t -= float(radius)
 
+        indices = index_is_monotonic(WMs[i].t)
+        WMs[i].t = WMs[i].t[indices]
+        WMs[i].data = WMs[i].data[indices]
+            
     # Create an instance of AsymptoticBondiData
     abd = AsymptoticBondiData(
-        time=WM_ref.t,
-        ell_max=WM_ref.ell_max,
+        time=WMs[list(WMs.keys())[0]].t,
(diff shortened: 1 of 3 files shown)


scri/SpEC/file_io/__init__.py
```diff
@@ -604,10 +604,14 @@ def create_abd_from_h5(
...
+        indices = index_is_monotonic(WMs[i].t)
+        WMs[i].t = WMs[i].t[indices]
+        WMs[i].data = WMs[i].data[indices]
+            
     # Create an instance of AsymptoticBondiData
     abd = AsymptoticBondiData(
-        time=WM_ref.t,
-        ell_max=WM_ref.ell_max,
+        time=WMs[list(WMs.keys())[0]].t,
+        ell_max=WMs[list(WMs.keys())[0]].ell_max,
```
moble, line 614: Why not still just use `WM_ref` here?  No copy was made, so `WM_ref` should have been changed in the same way, right?

keefemitman: Good point; reverted back!

keefemitman:
@moble not sure why the latest commit is causing `test_supertranslation_inverses` to fail... none of the code changed is relevant to that function... is something I should investigate?

moble:
I just made a commit that I thought should solve it, but I now realize won't do anything at all.  In any case, this is possibly a bad test, and the failure certainly unrelated to this commit, so I'll just merge this and deal with any failures later.  (Maybe I'll just disable the test if it keeps happening.)

PR #100, opened 2024-12-12
extrapolation: Print the error for the mode data validation

unkaktus:
As simple as that, there was a missing print statement for this validation case.


moble:
Thanks!

PR #101, opened 2025-03-06
Adding slicing to the abd data

akhairna:
Hi @moble, @duetosymmetry, and @keefemitman. I added the slicing option to the abd data which would be useful for my project. Please review the PR and let me know if there are any comments.

PR #103, opened 2025-04-03
Avoid unneeded interpolation in map_to_superrest_frame

akhairna:
Hi @moble & @duetosymmetry. I have replaced the interpolation with slicing in the `map_to_superrest_frame` function. Also, I spotted two typos in the description that I changed. This branch is rebased on top of the previous `slicing-abd` branch. Please review the PR.

### DIFF
--- a/scri/asymptotic_bondi_data/__init__.py
+++ b/scri/asymptotic_bondi_data/__init__.py
@@ -175,6 +175,38 @@ class AsymptoticBondiData:
         self._psi0[:] = psi0prm
         return self.psi0
 
+    #Slicing
+    def __getitem__(self, key):
+        """
+        Extract time slices of the asymptotic Bondi data efficiently.
+        """
+        # If key is a valid time slice or index, extract the corresponding
+        # sliced data
+        
+        if not isinstance(key, (slice, int)):
+            raise ValueError(f"Invalid key `{key}` of type `{type(key)}`.")                 
+
+        import copy
+        import functools    
+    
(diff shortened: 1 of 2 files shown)


duetosymmetry:
This PR is dependent on #101 (@moble, do you have a label?)

duetosymmetry:
@akhairna a better name for this PR would be "Avoid unneeded interpolation in map_to_superrest_frame"

duetosymmetry:
@keefemitman, Aniket found that the parameter `fix_time_phase_freedom` is just not used in `map_to_superrest_frame`. Was it supposed to be used for something? Instead of Aniket's documentation fix, we could just remove the parameter...

keefemitman:
Ah no this is suppose to call `sxs.waveforms.alignment.align2d`, but it looks like I never submitted a PR for this change which I had made locally. I'll submit a PR to do this tomorrow.

moble:
This looks like it *replaces* #101, no?

akhairna:
Yes, there would be some conflicts. I will be rebasing my commit on top of @keefemitman 's PR. Then my PR will be ready to review.

duetosymmetry:
#101 is not Keefe's PR, it is Aniket's earlier PR. The first two commits of this PR are the two commits from #101 (Aniket's original commit, plug Mike's merge commit). If #101 is merged, then this PR is atomic, just getting rid of interpolation when slicing could be used. @moble up to you if you want to review/merge them separately, or treat this as replacing #101.

PR #118, opened 2026-07-07
Add spectre versioning to abd metadata

keefemitman:
This changes the pipeline for the `"SpECTRECCE_v1"` pathway in `create_abd_from_h5` to extract metadata information from SpECTRE CCE's output .h5 file and append it to the metadata of the returned abd object as `version_info_update`. This can then be supplied to output waveforms, like the strain, by using `sxs.waveforms.rotating_paired_diff_multishuffle_bzip2.save`'s option `version_info_update`.

An example of what this additional information looks like as a part of the `version_info` is the following:

```
"spectre_cce": {
      "spectre_version": "2026.06.30",
      "spectre_git_branch": "develop",
      "spectre_git_revision": "e4ce00c0f",
      "spectre_link_date": "2026-07-06T14:49:32+00:00",
      "spectre_build_type": "Release",
      "spectre_build_information": "#\n# File created on Tue Jul  7 08:59:09 2026\n# SpECTRE Build Information:\n# Version:                      2026.06.30\n# Compiled on host:             mbot.cac.corne\
ll.edu\n# Compiled in directory:        /home/fs01/kem343/Codes/SpECTRE/build\n# Source directory is:          /home/fs01/kem343/Codes/SpECTRE\n# Compiled on git branch:       develop\n# Compiled on git \
revision:     e4ce00c0f\n# Linked on:                    2026-07-06T14:49:32+00:00\n# Build type:                   Release\n#",
      "spectre_cce_output_file": "CharacteristicExtractUnusedFile_R0100.h5",
      "spectre_cce_subfile": "SpectreR0100.cce",
      "spectre_header_source": "cce_subfile_header.hdr",
      "spectre_cce_sxs_format": "SpECTRE_CCE_v1",
      "spectre_cce_h5_object_version": 0,
      "spectre_h5_source_archive_location": "/src.tar.gz"
    }
```


scri/SpEC/file_io/__init__.py
```diff
@@ -528,28 +655,41 @@ def create_abd_from_h5(
...
-                    data_label_suffix = ".dat"
+            else:
+                cce_key = "Cce"
+                data_label_suffix = ".dat"
 
             cce = f[cce_key]
+
+            spectre_cce_version_info_update = (
+                _spectre_cce_version_info_update_from_h5(f, cce_key, file_name)
+            )
```
moble: I'm just a little nervous about assuming that this pretty complicated function that makes a lot of nested assumptions about where things are in files will always work, even for some old files people have lying around.  Should you wrap this in a `try` block?

scri/SpEC/file_io/__init__.py
```diff
@@ -528,28 +655,41 @@ def create_abd_from_h5(
                     radius = cce_key.split("R")[1][:4]
                     data_label_suffix = ""
                     break
-                else:
-                    cce_key = "Cce"
-                    data_label_suffix = ".dat"
+            else:
+                cce_key = "Cce"
+                data_label_suffix = ".dat"
```
duetosymmetry, line 660: Why did the indentation change here?

moble: TIL: This `else` block fires whenever `break` did not fire in the `for` loop — so if there are not items to iterate over *or* if the clause inside the loop never succeeded.  So this makes sure that these variables will always be defined.  LGTM.

moble: Actually, @keefemitman I guess this should get a comment explaining that point, since it was not obvious to either of us.

keefemitman: This has been added in the latest commit.

scri/SpEC/file_io/__init__.py
keefemitman: This has been added in the latest commit.

